PR #7, opened 2016-01-16
Add socks server

bitcoinfees:
Add the option to start a socks5 server using

`chisel server --socks PORT`

Currently no authentication method support.


jpillora:
chisel now has SOCKS5 support in `1.2.0`, please update, see `chisel server --help`, test it out and let me know if there are any issues

PR #10, opened 2016-03-19
Update client.go

i1v1s:


jpillora:
It'd be better to make a `--max-backoff` command-line option

i1v1s:
Hi.Check you dev@jpillora.com

PR #14, opened 2016-04-07
Add proxy support

chromicant:
Adds support for chisel client to talk over a HTTP proxy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+chisel.exe
--- a/client/client.go
+++ b/client/client.go
@@ -1,9 +1,12 @@
 package chclient
 
 import (
+	"crypto/tls"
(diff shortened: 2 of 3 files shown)


neutrinus:
Is there any reason why it's still not merged? I would like to use chisel with proxy... Thank you.

jpillora:
Will try to get this merged soon, just need to find time to test

gonsalez-ru:
Yes, issue is actual.

jpillora:
chisel now has HTTP CONNECT support in `1.2.0`, please update, see `chisel client --help`, test it out and let me know if there are any issues

instead of implementing a custom HTTP CONNECT client like this PR does, I swapped the websockets package to use gorilla which has proxy support built in

PR #16, opened 2016-06-09
Fix host is not being passed

### LINKED ISSUE #15: Problem in OpenShift
I'm trying to run chisel server on an OpenShift application but it fails. I think this is because Apache is listening on port 8080 by default (not sure).

Command:

```
./chisel server --host $OPENSHIFT_GO_IP --port $OPENSHIFT_GO_PORT --proxy http://example.com
```

Error:

`bind: address already in use`

KazeFlame:
Fix #15 too.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -89,14 +89,14 @@ func NewServer(config *Config) (*Server, error) {
 	return s, nil
 }
 
-func (s *Server) Run(host, port string) error {
+func (s *Server) Run(host string, port string) error {
 	if err := s.Start(host, port); err != nil {
 		return err
 	}
 	return s.Wait()
 }
 
-func (s *Server) Start(host, port string) error {
+func (s *Server) Start(host string, port string) error {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -106,7 +106,7 @@ func (s *Server) Start(host, port string) error {
 	}
 	s.Infof("Listening on %s...", port)
 
-	return s.httpServer.GoListenAndServe(":"+port, http.HandlerFunc(s.handleHTTP))
+	return s.httpServer.GoListenAndServe(host+":"+port, http.HandlerFunc(s.handleHTTP))
```
jpillora: Changes to lines 92,99 are unnecessary, though this looks good to me!

KazeFlame:
Removed unnecessary changes. Sorry I'm new to Go. I'm still learning the language.

jpillora:
No worries, thanks for the PR :)

PR #27, opened 2017-04-07
Client ssl support

cicavey:
I was excited to see your addition of socks5 and port to gorilla. I had already ported my fork to gorrilla and was working socks5. Your socks5 solution is much more elegant.

Additionally, I added support for client ssl. I like to secure my chisel endpoint with two way ssl. I imagine a way to specific a passphrase for a properly protected key vs my unencrypted key would make a nice addition. I'm happy to go down this path if you want.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,6 +1,7 @@
 package chclient
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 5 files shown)


jpillora:
Thanks for the PR. I was thinking of adding TLS, though decided not to for now.

Just a side note, see [tls.Config](https://golang.org/pkg/crypto/tls/#Config), `InsecureSkipVerify` disables all checks, not just parent certificates. With this enabled, TLS will encrypt though provide no authentication, so the only real point of doing this would be to pass through a TLS-only firewall. If we wanted to go down this path, we may as well generate a TLS certificate and hold it in memory (like I'm doing for the SSH certificates).

Put my thoughts on adding TLS here https://github.com/jpillora/chisel/issues/28

Issue #39, opened 2017-07-28
[RFC] Add conn pool for client


netroby:
To gain better connect performance, can we add conn pool for client? 

The current client, will dial remote when need.
It's ok, but not very fast.
If there be conn pool. it will gain better performance.

PR #40, opened 2017-07-28
Add conn pool. to gain better performance

netroby:


netroby:
see #39 

ping @jpillora

PR #41, opened 2017-07-28
Add conn pool support, gain better performance 

netroby:
See #39 
ping @jpillora

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -18,9 +18,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


jpillora:
Thanks for the PR! Are you able to provide any metrics for reconnection times, with and without this change?

AlekSi:
@netroby ping

jpillora:
Happy to look at this again with the metrics. Please reopen when ready

Issue #45, opened 2017-10-23
[Feature request] Allow client to config retry duration.


iineva:
Client was automatic add duration when connect fail.
But some times server is down at a long time, client can't retry in a few seconds.
Restart client is only i can do.

jpillora:
Implemented in 1.2.4

PR #47, opened 2017-10-26
Bug fix and add --max-retry-interval option.

### LINKED ISSUE #48: [BUG] Server side reboot client side can't reconnect.

iineva:
* Solve #45
* Fix #48

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -176,6 +176,7 @@ func (c *Client) loop() {
 		d := websocket.Dialer{
 			ReadBufferSize:  1024,
 			WriteBufferSize: 1024,
+			HandshakeTimeout: b.Duration(),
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -171,6 +176,7 @@ func (c *Client) loop() {
 		d := websocket.Dialer{
 			ReadBufferSize:  1024,
 			WriteBufferSize: 1024,
+			HandshakeTimeout: b.Duration(),
```
jpillora: What duration is this intended to be? `b.Duration` increments the backoff timer, though it's already being incremented when there's an error.

iineva: I use the same backoff timer for convenient. And logic is ok. Should we add another backoff timer to control `HandshakeTimeout`?

jpillora:
Thanks! See commits for changes

PR #67, opened 2018-09-21
use go modules

qiangli:
add gitignore
remove vendor folder
work around crypto/ssh handshake issue by locking to an old commit


jpillora:
Thanks! Hopefully will find some time soon to test

jpillora:
Hi, sorry for the delay! I haven't been able to test yet, though thought I'd check first: what was the crypto/ssh handshake issue?

qiangli:
It could be related to https://github.com/golang/go/issues/21941
I never had time to debug deeper as it works with a lower commit. Here are the steps to reproduce:
1. check out this pull request anywhere on your local machine (as it uses modules and does not need to reside in $GOPATH/src)
2. upgrade crypto/ssh in go.mod with: go get golang.org/x/crypto, you should see changes similar to the following:
-       golang.org/x/crypto v0.0.0-_20170825_220121-81e90905daef
+       golang.org/x/crypto v0.0.0-**20181015**023909-0c41d7ab0a0e
3. do a build: go build and run 
./chisel server
./chisel client -v  http://localhost:8080 localhost:6022:localhost:22
(client will hang at:  client: Handshaking...)

and
ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p 6022 $USER@localhost
will fail with the following error:
ssh_exchange_identification: Connection closed by remote host

jpillora:
Interesting I wonder if it's related to

https://stackoverflow.com/a/44269276

jpillora:
Oh wait, I think chisel sets the host key callback. Will investigate

jpillora:
Found it

https://github.com/golang/crypto/commit/5f55bce93ad2c89f411e009659bb1fd83da36e7b#diff-97976acd222484d4e11fa2a89268d611R354

Now, the SSH version string must contain `SSH-` otherwise handshake hangs

jpillora:
Did a full re vendor to get all the updates. So I didn't end up using your commit, though thanks for debugging and the PR! Checkout the 1.2.4 release!

qiangli:
Cool, now that you have identified the root problem. Thanks for your great work!

Issue #29, opened 2017-04-11
[Feature request] Reverse tunnel support


penguinpowernz:
It would be awesome to see this offer reverse port forwards as well as local port forwards.

penguinpowernz:
How to extend the remote spec to signal remote port forward? Prefix with an R? Make the centre colon an equals sign?

jpillora:
So this would make the server listen/accept connections and the client terminate them? I have another plan to allow clients to make connections https://github.com/jpillora/chisel/issues/9 though since it's quite a large change I decided not to include it in the `1.2` release

penguinpowernz:
Oh I'm thinking SSH protocol... perhaps I misunderstood the SSH related lines in your code.

Internally do you use SSH to do the local port forwards?  If so we could simply use that to enact remote port forwards.

jpillora:
No, internally, I use the SSH protocol. The SSH program implements it, so chisel would also need to implement it.

abraiante:
I second this. Initially I thought this was what Chisel did, but when setting it up I noticed it was the other way around.

My use case:

I have a local (like in the client) HTTP service, that I want to expose to the internet. Chisel is running on the server and I would to set-up a reverse tunnel, so issuing for instance, in the server a 

`
wget http://localhost:8080
`

would open a connection to a port in the client where the HTTP service is offered.

The reverse tunnel would be set-up when Chisel clients connects to the server, and as @penguinpowernz mentioned it could be described in the command-line with something like the thing SSH client uses:

```
-L [bind_address:]port:host:hostport
             Specifies that the given port on the local (client) host is to be forwarded to the given host and port on the remote side.  This works by allocating a socket
             to listen to port on the local side, optionally bound to the specified bind_address.  Whenever a connection is made to this port, the connection is forwarded
             over the secure channel, and a connection is made to host port hostport from the remote machine.

 -R [bind_address:]port:host:hostport
             Specifies that the given port on the remote (server) host is to be forwarded to the given host and port on the local side.  This works by allocating a socket
             to listen to port on the remote side, and whenever a connection is made to this port, the connection is forwarded over the secure channel, and a connection is
             made to host port hostport from the local machine.
```

This would make chisel one of the most versatile hole-punching firewall tools I know.

vizv:
> Initially I thought this was what Chisel did, but when setting it up I noticed it was the other way around.

That's exactly what I thought... After I setup the server, I found it's impossible to do the reverse tunneling.

[...]

PR #72, opened 2018-12-24
add reverse port forwarding

sunshineco:
This patch series implements reverse port forwarding (sharing client ports to the server) which complements existing forwarding (sharing server ports to the client). This is analogous to how OpenSSH's `-R` complements `-L` port forwarding.

Reverse port forwarding remotes are specified as `R:<local-interface>:<local-port>:<remote-host>:<remote-port>` and are subject to `--authfile` restrictions, just like normal remotes. A new server-side `--reverse` option controls whether `R:` remotes are accepted by the server (they are rejected by default).

I implemented these changes in isolation before ever seeing #29, which requests reverse port forwarding, or [github.com/mcbernie/chisel](https://github.com/mcbernie/chisel) which forks `chisel` to add `@client` endpoint notation, neither of which has seen any activity in over a year. The reverse port forwarding introduced by this series, on the other hand, works out-of-the-box, is available today, and is implemented with minimally invasive changes.


matti:
Add also support for dynamic remote port allocation like in SSH? (eg. port 0 and then print out the dynamically allocated port in log)

sunshineco:
> Add also support for dynamic remote port allocation like in SSH? (eg. port 0 and then print out the dynamically allocated port in log)

It looks like this could be done without major surgery since the server spins up the reverse-port-forwarding proxies before shipping its reply back to the client upon connection establishment, which means that the server can report all the dynamically-allocated ports to the client without changing the protocol (i.e. adding another client/server exchange). One small issue is that the client presently interprets any payload shipped back as an error and aborts the connection attempt. This will need to change to allow the server to ship back dynamically-allocated port numbers as non-error payload. Not a big deal, but something to take into consideration.

Anyhow, such functionality can be implemented atop the current pull request rather than re-rolling (though, I may take a stab at re-rolling this one if it is not merged in a timely fashion).

jpillora:
Merged :) Thanks @sunshineco, I do indeed prefer this PR over the other ones for its simplicity. There's still room to add `client --name <id>` and `@client-id` remotes, though I'd want a good chunk of time to merge existing implementations / rewrite it.

RE: Leaking go routine: So you run chisel via the Go API, and call `client.Start` a lot?

sunshineco:
> RE: Leaking go routine: So you run chisel via the Go API, and call `client.Start` a lot?

No, not yet, but I had it in mind to use `chisel` as a library, was glad to see it was already structured as such (with `main.go` just a thin wrapper over the _libified_ client and server code), so when I saw the leak, I plugged it.

tijme:
I would love to see dynamic reverse port forwarding. Maybe I'll take a shot at it myself. 😄 

Nice pull request @sunshineco!

Issue #52, opened 2017-12-17
[Feature request] Option to custom HTTP header in client


retred:
Hi.
Add please option custom HTTP header in client.These settings help many bypass the firewall.
User-Agent:CustomAgent
Host:CustomHost
X-Forwarded-Host:CustomHost

The current settings for bypassing the firewall are very bad -
User-Agent:Go-http-client/1.1

Please add this option. 
I think to you many will say thank you. 
And your program will be the best.

jpillora:
Interesting, are you finding that `Go-http-client/1.1` is blocked?

aus:
If you are using a HTTP proxy in chisel client, the User-Agent `Go-http-client/1.1` will still be exposed during the initial CONNECT request to the proxy. Unfortunately, it seems that it is not possible to customize the CONNECT request header with gorilla/websockets. I've opened and issue [here](https://github.com/gorilla/websocket/issues/479).

eliezedeck:
Not only those, but possibly others, possibilities are endless. Please integrate this feature. Thanks.

ingnelson:
Will you add Custom HTTP header , sir ? This is important to bypass firewall .

Issue #35, opened 2017-06-10
[Feature request] Add option to customize the "Host" field in HTTP header in client


t123yh:
Some firewalls will check the `Host` header in the request and see if it's in the white list. I wish there could be a command-line option to customize this field (and possibly, arbitrary header fields).

undirectlookable:
same as #33

matti:
this has been implemented and this can be closed now?

PR #76, opened 2019-01-19
Add optional HOST header

nathan0:
Will add the optional `--hostname <host>` argument to the client. Will close issues [#52](https://github.com/jpillora/chisel/issues/52) and [#35](https://github.com/jpillora/chisel/issues/35)


jpillora:
Thanks!

Issue #74, opened 2019-01-07
authfile not working


enzokwai:
Issue
I am not able to start chisel server when using authfile option, it output fatial error and exit.

Everything is normal when not using authfile

I am try to use sample authfile in github still no luck

Context
λ chisel server --port 8020 --authfile users.json
2019/01/04 18:35:28 server: users: Loading the configuraion from: users.json
fatal error: sync: Unlock of unlocked RWMutex

goroutine 1 [running]:
runtime.throw(0x845eb3, 0x20)
/app/.heroku/go/src/runtime/panic.go:608 +0x79 fp=0xc00007fb08 sp=0xc00007fad8 pc=0x42d4b9
sync.throw(0x845eb3, 0x20)
/app/.heroku/go/src/runtime/panic.go:594 +0x3c fp=0xc00007fb28 sp=0xc00007fb08 pc=0x42d42c
sync.(*RWMutex).Unlock(0xc000005500)
/app/.heroku/go/src/sync/rwmutex.go:124 +0xa8 fp=0xc00007fb58 sp=0xc00007fb28 pc=0x46e1c8
github.com/jpillora/chisel/share.(*UserIndex).loadUserIndex(0xc0000051a0, 0x846157, 0x21)
/tmp/cloudgox/080f9a4f2cc2/gopath/src/github.com/jpillora/chisel/share/users.go:151 +0x702 fp=0xc00007fcb0 sp=0xc00007fb58 pc=0x72a262
github.com/jpillora/chisel/share.(*UserIndex).LoadUsers(0xc0000051a0, 0xc000018180, 0xa, 0x0, 0xcb)
/tmp/cloudgox/080f9a4f2cc2/gopath/src/github.com/jpillora/chisel/share/users.go:78 +0xdc fp=0xc00007fd10 sp=0xc00007fcb0 pc=0x7299bc
github.com/jpillora/chisel/server.NewServer(0xc00007fec0, 0xa, 0x0, 0x0)
/tmp/cloudgox/080f9a4f2cc2/gopath/src/github.com/jpillora/chisel/server/server.go:63 +0xa9a fp=0xc00007fdf8 sp=0xc00007fd10 pc=0x76468a
main.server(0xc000084020, 0x4, 0x6)
/tmp/cloudgox/080f9a4f2cc2/gopath/src/github.com/jpillora/chisel/main.go:174 +0x4d4 fp=0xc00007ff18 sp=0xc00007fdf8 pc=0x766544
main.main()
[...]

PR #77, opened 2019-01-22
Fixing issue #74

patrick-othmer:
Fixing issue #74.

Lock destroyed with u.Users = users. Lock not needed if we directly use Users.


jpillora:
Thanks! Sorry for the slow merge. Will make a new release now.

jpillora:
Released 1.3.1

PR #79, opened 2019-02-09
 add support for connecting to chisel server via socks5 proxy in client 

Meteorite:
Make it possible to connect client to server via socks5 proxy (in addition to already support http CONNECT proxy):
```
chisel client -proxy socks://socks5-proxy:1080 https://chisel-server socks
2019/02/09 23:39:11 client: Connecting to wss://chisel-server:443 via socks://socks5-proxy:1080
2019/02/09 23:39:11 client: proxy#1:0.0.0.0:1080=>socks:1080: Listening
2019/02/09 23:39:13 client: Fingerprint 5c:3c:e7:52:22:9b:09:ee:f8:87:4e:be:36:95:9c:0b
2019/02/09 23:39:13 client: Connected (Latency 91.2ms)
```

Authentication is also supported:
```
chisel client -proxy socks://admin:bigsecret@localhost:1100 https://chisel-server socks
```

As a side effect, it is now possible to chain chisel instances via each other :)


jpillora:
LGTM 👍 thanks for the PR, just need to get around to testing this and I'll get it merged :)

aus:
Nice PR! 

One suggestion: maybe we should model the proxy URIs like curl. `socks://` would be better specified as `socks5://`. Then maybe in the future we could support additional protocols such as `socks4://`, `socks4a://` or `socks5h://`.

Meteorite:
I used golang.org/x/net/proxy implementation of socks5 client, that never resolves domains on its side and always passes them to socks server (see [here](https://github.com/golang/net/blob/74dc4d7220e7acc4e100824340f3e66577424772/internal/socks/client.go#L86-L103)). So technically it works like `socks5h://`, which is a good default to my mind :)
But for now you can already use either `socks://` or `socks5://` or `socks5h://` or everything, starting with `socks` (see [here](https://github.com/jpillora/chisel/pull/79/commits/7497a3acca7dc21f29b686ade96346e2fbe47953#diff-1cd41a2611da49ce63630d5a51bd82c5R199)). It will work as `socks5h://` regardless of what exactly you specified though...

I agree, that may be it is better to allow only `socks://` and `socks5h://` for now, making `socks5h://` a default for `socks://`.

jpillora:
Merged! Sorry for the slight 😀delay

PR #84, opened 2019-03-07
add ability to skip tls certificate verification

Meteorite:
Add `--skip-tls-verification` option for client to allow connecting to front TLS-terminating servers with self signed certificates:

```
chisel client --skip-tls-verification https://chisel-server socks
```

Resolves #34
Built on top of #79

PR #85, opened 2019-03-08
add support for connecting to user endpoint via socks5 proxy in chisel server

### LINKED ISSUE #81: Server support upstream socks5 proxy
In some case there need a domain or ip rule based software for dispatch traffic on server side
Such as https://github.com/Dreamacro/clash Config:
```
Rule:
- DOMAIN-SUFFIX,google.com,auto
- DOMAIN-KEYWORD,google,auto
- DOMAIN,google.com,auto
- DOMAIN-SUFFIX,ad.com,REJECT
- IP-CIDR,127.0.0.0/8,DIRECT
- SOURCE-IP-CIDR,192.168.1.201/32,DIRECT
```
I hope chisel server be able to forward outgoing traffic to another socks5 server

Meteorite:
Make it possible to specify socks5 proxy server, which will be used for all outbound connections from chisel server to different user endpoints, to which connected clients want to connect to:
```
chisel server --upstream-proxy socks5-proxy:1080 --socks5
2019/03/08 17:13:16 server: SOCKS5 server enabled
2019/03/08 17:13:16 server: Fingerprint 48:e9:20:fd:2e:48:96:d6:8a:8b:6c:10:46:1f:58:7b
2019/03/08 17:13:16 server: Upstream proxy: socks5://socks5-proxy:1080
2019/03/08 17:13:16 server: Listening on 0.0.0.0:8080...
```
Authentication is also supported:
```
chisel server --upstream-proxy admin:bigsecret@localhost:1100
```
Resolves #81 
Built on top of #79 and #84


vetch101:
Is there a way to get the Chisel socks5 proxy to listen on any IP other than 127.0.0.1?

jpillora:
It doesn't listen. It creates an internal chisel endpoint to a socks5
connection handler.

vetch101:
Thanks for the reply.

I'm really enjoying using chisel, it's great! Thank you!

I've got a setup where I'm using it with a deployment where I'm using it for log forwarding over the tunnel.

The situation is:

application node (filebeat) log forwarder --> remote chisel client -----------> 
----> local chisel server --> local log (logstash) pipeline 

This works great when the tunnel is up and the logstash server is up, or when the tunnel is down, but when the logstash server is down and the tunnel is up, the log forwarder floods the network with logs as it sees the chisel connection, but (obviously) can't reach the end point.

I thought that I might be able to get it to work by authenticating the connection with SSL certs, and the filebeat log forwarder backs off the tunnel when it's up and the logstash is down, but it also backs off when the logstash server is up as it can't authenticate...

I tried using it with socks5 proxy - and it all works fine with no SSL (although it has the same problem with the log forwarder flooding the network when it can see chisel, but not the logstash endpoint)... and when I try with SSL, I get the same outcome as with the other situation...

That's why I thought it might be helpful to be able to listen on a non-loopback address; but from your reply, it sounds like that isn't possible.

Are there any other ways that you could suggest that might be viable?

Many thanks!

jpillora:
I'd suggest having your own logic do a health check against remote log
server, and toggle a switch to stop sending logs (or do whatever happens
when the log endpoint, chisel in this case, goes down)

vetch101:
Makes sense - thank you!

matti:
any change to get this merged?

PR #114, opened 2019-09-21
Fix typo in README.md

johschmitz:


jpillora:
Thanks looks like this now fixed

PR #90, opened 2019-03-28
Add support for custom HTTP headers

AkeemMcLennon:
@jpillora This pull request adds support for custom http headers in the client request (beyond host). This can be useful for getting behind reverse proxies that require authentication headers for processing the request.

I also refactored the `--hostname` parameter to use the custom header logic and added a few unit tests for the new functionality. 

The new tests can be run through `go test ./...`

PR #124, opened 2019-10-30
Feature/custom headers support

alfonso-presa:
This allows to add custom Http/s headers. In our case, we need this because we have the server behind a basic auth secured reverse proxy.


jpillora:
Thanks for your work, though #90 was merged into `1.5.0` so I think I can close this

PR #116, opened 2019-09-25
add support for skipping tls certificate verification

alfonso-presa:
cherry-picked from: https://github.com/jpillora/chisel/pull/84
credits: https://github.com/Meteorite

closes: #34

PR #129, opened 2019-11-25
TLS serving for server and CA verification to client

### LINKED ISSUE #34: [Feature request] Option to Ignore TLS certificate verification
(Sorry for my poor English)
Chisel is an amazing tool! I want to use chisel behind Nginx.
Becasuse the firewall of my company only allow port 443 use HTTP CONNECT  command. 
I want to use port 443 for my home server for HTTPS website, and chisel too. 
So I tried to config Nginx to redirect HTTPS request by different domain.

[Client Command]
chisel_windows_amd64.exe client --proxy http://company-proxy:port https://home-server:443 0.0.0.0:33389:192.168.1.102:3389

2017/05/26 11:34:54 client: Retrying in 100ms...

mvladev:
This PR adds a new optional flag to the client `--ca-file` which can be used to verify certificate of the server when using `wss` or `https`.

The server is also capable of serving TLS via `--tls-cert-file` and `--tls-keyfile` flags.

Fixes #34 and it's a more secure alternative of #116.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ $ chisel server --help
     access, in the form of <user:pass>. This is equivalent to creating an
     authfile with {"<user:pass>": [""]}.
 
+    --tls-cert-file, An optional file containing the default x509 Certificate
+    for HTTPS. (CA cert, if any, concatenated after server cert).
+
+    --tls-key-file, An optional file containing the default x509 private key
+    matching "--tls-cert-file".
+
     --proxy, Specifies another HTTP server to proxy requests to when
     chisel receives a normal HTTP request. Useful for hiding chisel in
     plain sight.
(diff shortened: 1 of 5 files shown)


mvladev:
After spending some more time looking/testing/writing at various tunnels

- https://github.com/ginuerzh/gost (various tunneling options)
- https://github.com/jpillora/chisel
- https://github.com/xtaci/kcptun (transport is UDP-based)

As of terms of performance, `chisel` was the fastest one + least latency. The architecture is simple:

overview

It creates logical TCP connections via SSH, but it uses only one TCP connection per client/Shoot Node.
With this approach a Chisel server (with appropriate replica count) can be deployed and it'll serve as a proxy into the cluster for those Nodes.

The only security requirement I've seen so far is the lack of TLS (the client uses fingerprints for auth), but I've opened a PR to [fix it](https://github.com/jpillora/chisel/pull/129).

One interesting point the fastest proxy (in terms of overhead/latency) is one I've written from scratch and it's based on the new [QUIC transport protocol](https://tools.ietf.org/html/draft-ietf-quic-transport-24). This protocol used as the transport protocol for HTTP3 and supports multiplexing, multiple streams and TLSv1.3. The proxy + client is only several hundreds LOC. The only downside is that it's UDP-based and support for UDP is lacking (different OpenStack deployments may not support it). Support for QUIC in Envoy is also [coming soon™](https://github.com/envoyproxy/envoy/issues/2557).

I'll create a new repo for the sidecar component and start writing a GEP.

jpillora:
See commit! Releasing this in `v1.7.0-rc7`

PR #130, opened 2019-12-15
Add encrypted password support

cagdasbas:
Currently supports two encryption method: sha256 and sha512. Can be added on compile time.

User can create a hashed password with `mkpasswd --method=sha-512`, Server will get the usernames and password as usual, client will supply a hashed password:
`chisel client --auth user:'$6$QYGHuvnm1c29b0cm$IzOCBgI39CrAphCfRHSwWkpYLGW0f7/gUUXyzjG6W6gjKdggZ3qh4x6CD7H9ixmyOHYwZaYqGdqJKSZpeSw9J1' `

Server will try to create a Crypter with the given password prefix ($5$ is sha256 and $6$ is sha512) and try to decrypt it. If a Crypter not available, it will try a plaintext authentication.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/jpillora/requestlog v0.0.0-20181015073026-df8817be5f82
 	github.com/jpillora/sizestr v0.0.0-20160130011556-e2ea2fa42fb9
 	github.com/tomasen/realip v0.0.0-20180522021738-f0c99a92ddce // indirect
+	github.com/tredoe/osutil v0.0.0-20191018075336-e272fdda81c8
 	golang.org/x/crypto v0.0.0-20181015023909-0c41d7ab0a0e
 	golang.org/x/net v0.0.0-20181017193950-04a2e542c03f // indirect
 	golang.org/x/sys v0.0.0-20181019160139-8e24a49d80f8 // indirect
--- a/go.sum
+++ b/go.sum
@@ -16,6 +16,9 @@ github.com/jpillora/sizestr v0.0.0-20160130011556-e2ea2fa42fb9 h1:0c9jcgBtHRtDU/
 github.com/jpillora/sizestr v0.0.0-20160130011556-e2ea2fa42fb9/go.mod h1:1ffp+CRe0eAwwRb0/BownUAjMBsmTLwgAvRbfj9dRwE=
 github.com/tomasen/realip v0.0.0-20180522021738-f0c99a92ddce h1:fb190+cK2Xz/dvi9Hv8eCYJYvIGUTN2/KLq1pT6CjEc=
 github.com/tomasen/realip v0.0.0-20180522021738-f0c99a92ddce/go.mod h1:o8v6yHRoik09Xen7gje4m9ERNah1d1PPsVq1VEx9vE4=
+github.com/tredoe/goutil v0.0.0-20161130132832-0a73aea41b0b/go.mod h1:dp4VPOLeEFYbsf1ikgd+uytWDnpCdMiTHMg6mh7hHuQ=
+github.com/tredoe/osutil v0.0.0-20191018075336-e272fdda81c8 h1:kKa/vDK8CCEnDLug5cfYRHZZJ2JKu4/wyHmjkzaOLk0=
+github.com/tredoe/osutil v0.0.0-20191018075336-e272fdda81c8/go.mod h1:w7hqLjZRokyWIpiEXWj6pXIHOg/2tSWSBsoYfdc9bjw=
 golang.org/x/crypto v0.0.0-20181015023909-0c41d7ab0a0e h1:IzypfodbhbnViNUO/MEh0FzCUooG97cIGfdggUrUSyU=
 golang.org/x/crypto v0.0.0-20181015023909-0c41d7ab0a0e/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/net v0.0.0-20181017193950-04a2e542c03f h1:4pRM7zYwpBjCnfA1jRmhItLxYJkaEnsmuAcRtA347DA=
--- a/server/server.go
+++ b/server/server.go
@@ -2,6 +2,8 @@ package chserver
 
 import (
 	"errors"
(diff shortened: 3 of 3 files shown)


scriptcoded:
This seems to have dangled for a while. Any plans to integrate this @jpillora? Is there anything I could to to help in the process?

jpillora:
Hey Malcolm

This one looks okay, though there’s a rare use case however which breaks
this: If someone is using a valid encoded hash as a password.

We could guess that it would be rare, and make this change anyway?

Also FYI, I’ve made a very large refactor over on the “udp” branch. So if
you are looking into sending a PR, base it off that. Even safer would be to
wait for it to get merged into master.

cagdasbas:
Hi Jaime,

I can check the plaintext match first if you think that case has to be handled.

I think rebasing after the merge would be safer indeed.

scriptcoded:
I can think of two other solutions

1. Hash the username and password together, resulting in a hash containing no colons. The downside would be that the username isn't visible in plain text.
2. Introduce some form of prefix. I've got three ideas on how this could be implemented.
    1. Use a prefix before the password, like this: `foo:sha256:$6$QYGHu...`. This would probably require any colon in the password to be escaped to prevent confusion with passwords beginning with `sha256:`
    2. Use a prefix before the password, but require plaintext passwords to use an empty field, like this `foo::bar` or `foo:plain:bar`. Perhaps this would only be necessary for passwords containing colons?
    3. Put the prefix before the username, resulting in something like `sha256:foo:$6$QYGHu...`. The issue here would be that you couldn't use sha256 as a username.

Apache takes a similar approach and uses a `{SHA}` prefix followed by a base64-encoded hash. https://httpd.apache.org/docs/2.4/misc/password_encryptions.html

cagdasbas:
There is already a prefix. $6$ means sha512 for example. It's a standard crypt(C) prefix: [https://man7.org/linux/man-pages/man3/crypt.3.html](https://man7.org/linux/man-pages/man3/crypt.3.html)

The crypt format defines an encrypted format as:
`$id$salt$encrypted`

I don't think we should hash the username together. Usually, the password is encrypted and username is plaintext. SSL transmission ensures the safety of the usernames. It's also harder to create. `mkpasswd --method=sha-512` will not show you what you are writing and typing `<username>:<password>` is harder if you are not seeing and people are used to typing only their passwords. It's unnecessarily complex.

jpillora:
Hmm yeah I think we opt for simplicity: we just do a prefix check ($6$) and
a string length check (exactly sha512 chars long), then treat it as a hash,
otherwise treat it as plaintext. And anyone who uses a hash as a password
will just have to update their clients.

scriptcoded:
@cagdasbas Sorry, that's me rushing through things. I totally missed the prefix. I took a closer read through your description, and I realize I've probably misunderstood the PR. You're aiming to hash the password at the client level to ensure safe delivery, not hashing the password at the server level for secure storage. Am I right? I think both are valid cases, and perhaps both should be considered.

jpillora:
It’s for safe storage, since safe delivery (encrypted and authenticated) is
guaranteed by the ssh protocol (and tls too if enabled).

It also makes it slightly harder to brute force. Incorrect password
attempts should be rate limited but they’re not yet.

cagdasbas:
@jpillora The encryption algorithm does that actually. `crypt.NewFromHash(string(password))` returns a null encryption algorithm if it cannot decrypt the given string. 

@scriptcoded No problem! Yes, you are right. We are running the client on client servers to create a secure tunnel to our Apache Guacamole server for remote support purposes (within their knowledge of course). That's why we don't want them to see the plaintext passwords in a config file. @jpillora The current changeset doesn't cover secure password storage on the server-side. I haven't changed the auth file load logic on the server-side. It just decrypts incoming auth password and compares with the stored plaintext password. Now you guys mentioned it, both make sense to me.

jpillora:
Oh right, i also misunderstood - the client is sending the hash.
Interesting. Yes both make sense. The tricky part is you can’t have a hash
on both sides...

jpillora:
Actually, if you generate a string random password on the server, is there any improvement in security by the client sending a hash?

I think the only benefit comes from safe storage...

cagdasbas:
Exactly :) Should there be two options? You can choose either client-side or server-side encrypted password storage. Is there an optimal way to handle this?

You are correct. There isn't in the short run. But my second PR will be on LDAP authentication. Couldn't find time to do anything about it, it was in my mind and I was going to create an issue about it to ask your opinion on that. In that situation, the client shouldn't see the password at all. Maybe that's a completely different and future problem that you don't want to handle right now. 

The one example I can think of is virt-who. It encrypts and stores the password but I think it encrypts with an SSL key stored in the same machine. Decrypts the password and sends the password plaintext. I'm not sure about the pipeline though.

jpillora:
Since there's no gain in security from hashing on the client side, I think it only makes sense to hash for secure storage on the server side for now.

Basically, hashing is an attempt to prevent password stealing. This works on the server because even if an attacker can steal the hash, they still don't know the password and can't impersonate the client. This is not true for client side hashing, if an attacker steals a client side hash, then they can impersonate the client - therefore a password is equivalent and the hash is not required.

The next use-case is brute force prevention, though I think IP rate limiting is the answer here.

scriptcoded:
@jpillora That sounds like a reasonable conclusion. Rate limiting should be in its own separate PR though, shouldn't it? And then @cagdasbas will have to decide if changing the scope of this PR is easy enough, or if raising a new one makes more sense.

cagdasbas:
Well, in that case, I'm closing this PR. I'll rebase the encrypted password support on UDP branch but for the server-side this time. I can implement it on the master branch if you prefer.

jpillora:
Thanks guys Will close this one and feel free to open another adding support for server-side password hashes

srueg:
@cagdasbas did you ever get around implementing the server-side hashing of passwords?
I might give it a go otherwise 👍

cagdasbas:
@srueg Couldn't find time to implement it to the server side because of all the things going on. Give it a go. Maybe you can check out my client side implementation on my fork.

pedroapero:
Hi, just saying but it would probably be easier to use ssh keys pairs instead of using password hashes as this is implemented in crypto/ssh.

Issue #153, opened 2020-04-17
mips and mipsle support


pexcn:
Hi jpillora:

Can chisel add mips and mipsle platform support?
like v2ray-core: https://github.com/v2ray/v2ray-core/releases

Thanks 🍻

jpillora:
Done https://github.com/jpillora/chisel/releases sorry forgot to post back here

pexcn:
Thank you :)

PR #154, opened 2020-04-25
release mips and mipsle version

pexcn:
Hi jpillora,

This commit should fix https://github.com/jpillora/chisel/issues/153, but I have not tested, so I am not sure if my config is correct, please review it.

My expectation is that it can add the following versions:
- mips-hardfloat
- mips-softfloat
- mipsle-hardfloat
- mipsle-softfloat

See more:
https://github.com/goreleaser/goreleaser/pull/1288

### DIFF
--- a/.github/goreleaser.yml
+++ b/.github/goreleaser.yml
@@ -12,9 +12,14 @@ builds:
       - amd64
       - arm
       - arm64
+      - mips
+      - mipsle
     goarm:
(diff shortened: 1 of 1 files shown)


jpillora:
https://github.com/jpillora/chisel/blob/master/.github/goreleaser.yml#L22-L36 this is done

PR #167, opened 2020-06-23
In stdio mode, chisel exits if the tunnel for stdio is closed.

BoleynSu:
Avoid checking if sshConn is nil by using channel and os.Exit(0)
when the stdio connection ends.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -41,7 +41,7 @@ type Client struct {
 	*chshare.Logger
 	config      *Config
 	sshConfig   *ssh.ClientConfig
-	sshConn     ssh.Conn
+	sshConn     chan ssh.Conn
 	proxyURL    *url.URL
 	server      string
 	running     bool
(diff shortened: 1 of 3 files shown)


BoleynSu:
1) `accept` will keep getting "No remote connection" without this PR.
2) This PR also make chisel exit if stdio connection ends.

```golang
func (p *TCPProxy) accept(src io.ReadWriteCloser) {
        defer src.Close()
        p.count++
        cid := p.count
        l := p.Fork("conn#%d", cid)
        l.Debugf("Open")
        sshConn := p.ssh()
        if sshConn == nil {
                l.Debugf("No remote connection")
                return
        }
        //ssh request for tcp connection for this proxy's remote
        dst, reqs, err := sshConn.OpenChannel("chisel", []byte(p.remote.Remote()))
        if err != nil {
                l.Infof("Stream error: %s", err)
                return
        }
        go ssh.DiscardRequests(reqs)
        //then pipe
        s, r := Pipe(src, dst)
        l.Debugf("Close (sent %s received %s)", sizestr.ToString(s), sizestr.ToString(r))
}
```

PR #178, opened 2020-07-31
Add ability to watch parent pid and exit if it no longer exists.

marve:
This PR adds the optional ability for chisel to exit automatically if the parent process no longer exists.

`chisel server --watch-parent` -or- `chisel client --watch-parent <some-remote>`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/jpillora/backoff v1.0.0
 	github.com/jpillora/requestlog v1.0.0
 	github.com/jpillora/sizestr v1.0.0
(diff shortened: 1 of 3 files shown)


jpillora:
Whats the use case for this? running proc in bg with: `chisel ..... &`?

marve:
Something like that, yes. I am using this flag to avoid zombie processes for a web application wrapper around chisel. Under nominal shutdown conditions there is no problem but in certain edges, specifically in the integration test suite for this application where SIGINT may not be sent to the parent process, it cannot terminate chisel and thus subsequent test runs fail.

PR #185, opened 2020-08-24
mTLS authentication feature.

wxia20:
1. Adding client cert and key to Chisel client.
2. Adding trusted CAs to Chisel server.
3. Adding TLS/mTLS e2e test case and test certs and keys.
4. Adding flag and help comments.


client/client.go
```diff
@@ -46,6 +46,8 @@ type Config struct {
 type TLSConfig struct {
 	SkipVerify bool
 	CA         string
+	MtlsCliCrt string
+	MtlsCliKey string
```
jpillora, line 50: Since this struct is `client.TLSConfig`, and a client certificate implies MTLS, these can be renamed to `Cert` and `Key`

client/client.go
```diff
@@ -117,6 +119,18 @@ func NewClient(c *Config) (*Client, error) {
 				tc.RootCAs = rootCAs
 			}
 		}
+		//Specify client cert and key pair
+		if c.TLS.MtlsCliCrt != "" && c.TLS.MtlsCliKey != "" {
+			tc.GetClientCertificate = func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
```
jpillora, line 124: Instead of using `GetClientCertificate` (which is called repeatedly) you should set `tc.Certificates` as an array wth only `c`.

main.go
```diff
@@ -157,6 +157,11 @@ var serverHelp = `
     You can modify this path by setting the CHISEL_LE_CACHE variable,
     or disable caching by setting this variable to "-". You can optionally
     provide a certificate notification email by setting CHISEL_LE_EMAIL.
+
+    --tls-mtls-ca-dir, for mTLS support only. Specify the path to load 
+    all self signed client certificates and trusted CAs in PEM-encoded 
+    format. For trusted CAs, any client certificates signed by these CAs 
+    consider to be allowed.
```
jpillora, line 164: Please update this to:

```
--tls-ca, a path to a PEM encoded CA certificate bundle used to validate client connections. The provided CA certificates will be used instead of the system roots. This is commonly used to implement mutual-TLS.
```

main.go
```diff
@@ -367,6 +373,12 @@ var clientHelp = `
     transport https (wss) connection. Chisel server's public key
     may be still verified (see --fingerprint) after inner connection
     is established.
+
+    --tls-mtls-clicrt, for mTLS support only. PEM-encoded client certificate
+    for mTLS authentication.
+
+    --tls-mtls-clikey, for mTLS support only. PEM-encoded client key for
+    mTLS authentication.
```
jpillora, line 381: Please update these to:

```
--key, a path to a PEM encoded private key used for client authentication (mutual-TLS).

--cert, a path to a PEM encoded certificate matching the provided private key. The certificate must have client authentication enabled (mutual-TLS).
```

server/server_listen.go
```diff
@@ -13,9 +15,10 @@ import (
 
 //TLSConfig enables configures TLS
 type TLSConfig struct {
-	Key     string
-	Cert    string
-	Domains []string
+	Key       string
+	Cert      string
+	Domains   []string
+	MtlsCaDir string
```
jpillora, line 21: Please rename this to `CA`

server/server_listen.go
```diff
@@ -82,13 +85,41 @@ func (s *Server) tlsLetsEncrypt(domains []string) *tls.Config {
...
 	if err != nil {
 		return nil, err
 	}
-	//return file based tls config using tls defaults
-	return &tls.Config{
-		Certificates: []tls.Certificate{c},
-	}, nil
+	//mTLS requires Client CAs
+	if clientCaPath != "" {
+		files, err := ioutil.ReadDir(clientCaPath)
```
jpillora, line 95: CAs can be encoded into bundles. So you can have multiple CAs concatenated in a single file.

server/server_listen.go
```diff
@@ -82,13 +85,41 @@ func (s *Server) tlsLetsEncrypt(domains []string) *tls.Config {
...
+			return nil, err
+		}
+		s.Infof("Looking for client CA files from %s", clientCaPath)
+		clientCAPool := x509.NewCertPool()
+		//add all cert files from path
+		for _, file := range files {
+			f := file.Name()
+			content, err := ioutil.ReadFile(filepath.Join(clientCaPath, f))
+			if err == nil {
+				if clientCAPool.AppendCertsFromPEM(content) {
```
jpillora, line 106: Instead of reading individual files, https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM this function supports CA bundles. So you only need to call this once on a single CA file.

server/server_listen.go
```diff
@@ -82,13 +85,41 @@ func (s *Server) tlsLetsEncrypt(domains []string) *tls.Config {
...
+					s.Infof("Add client CA file %s", f)
+				} else {
+					s.Errorf("Fail to add client CA file %s", f)
```
jpillora, line 109: I think we should `return nil, err` not just print

server/server_listen.go
```diff
@@ -82,13 +85,41 @@ func (s *Server) tlsLetsEncrypt(domains []string) *tls.Config {
...
+				}
+			}
+		}
+		//return file based tls config with client CAs and cert verification enabled
+		return &tls.Config{
+			Certificates: []tls.Certificate{c},
+			ClientCAs:    clientCAPool,
+			ClientAuth:   tls.RequireAndVerifyClientCert,
+		}, nil
+	} else {
```
jpillora, line 119: instead of having `else` with 2 cases. We can make this 1 case by initialising `&tls.Config{}` at the top, and then only setting `ClientCAs` and `ClientAuth` in the CA block

test/e2e/tls/client-ca/server.crt
```diff
@@ -0,0 +1,16 @@
+-----BEGIN CERTIFICATE-----
```
jpillora, line 1: Self signed certs are good for tests, can you tell me when they expire? Sorry cant decode them at the moment

jpillora (commented):
Thanks for the PR! Everything is mostly in the right place, just a bit renaming and other minor edits

test/e2e/tls/client-ca/server.crt
wxia20: Both server and client certs are valid in 10 years. 2030.

Screen Shot 2020-08-25 at 11 41 06 AM
Screen Shot 2020-08-25 at 11 42 27 AM

client/client.go
wxia20: I try to differentiate from server cert and key by name. I can rename it put comment.

client/client.go
wxia20: Will try.

main.go
wxia20: 👍

main.go
wxia20: 👍

server/server_listen.go
wxia20: Sure, as long as developers not confuse this CA with server cert CA.

server/server_listen.go
wxia20: I thought about this when I work on this. In reality, most of the users will put individual client certs (self signed) over here instead of one trusted CA. If we only allow one CA bundle, that means users have to maintain a combined CA bundle by themselves, of which mostly likely they don't know how to do it. Even they can handle it, adding and removing certs will be a hassle. Please give a thought on this.

server/server_listen.go
wxia20: please see my comment above.

server/server_listen.go
wxia20: 👍

server/server_listen.go
wxia20: Will try.

client/client.go
jpillora: Like I mentioned, the struct `client.TLSConfig` differentiates it from `server.TLSConfig` already

server/server_listen.go
jpillora: If I start chisel, explicitly pointing at one or more CAs, I want to use them or not run at all. Definitely should exit 👍

test/e2e/tls/client-ca/server.crt
jpillora: 👍

server/server_listen.go
jpillora: Okay, fair enough. Though it should also support single files. Can `os.Stat` and then `if IsDir()` do your logic, `else` load single file.

client/client.go
wxia20: 👌

server/server_listen.go
wxia20: sounds good

server/server_listen.go
wxia20: 👌

wxia20:
@jpillora I submit 1121eb2, it's weird that CI works on Ubuntu and Mac but failed on windows. Do you have any clue, I use Mac locally and it works all good. Any clue?

PR #187, opened 2020-09-04
SFTP Support

aus:
I had an interesting use case for chisel where I needed to access SFTP. This is normally possible simply by tunneling an SSH connection through chisel. However since chisel already uses SSH channels, this results in double encapsulation of SSH. This PR adds support for an internal SFTP server, similar to how the SOCKS remote works. The SFTP code makes use of [github.com/pkg/sftp](https://github.com/pkg/sftp) which is a decent implementation, but Windows support is a bit wonky (working directory commands like `cd` or `pwd` do not work, but file operations like `get`, `put` and `ls` work fine). 

I should also note that the `sftp` remote works a bit different than users might expect. It might help to understand a bit of background about how SFTP works. SFTP clients are basically just wrappers around SSH. When you do a `sftp user@host`, you are really invoking SSH and executing the sftp subsystem on the remote host. Your SFTP client and the server's SFTP subsystem establish a pipe and the connection is ready for commands.

However, chisel has already established a SSH channel. To avoid the double SSH encapsulation, we need to inform our SFTP client to use an external program for establishing an encrypted connection. `sftp` passes SSH options to this program and expects it to establish a secure connection before establishing a pipe with the SFTP server. chisel has already done this for us, so we can simply use netcat to establish a connection directly to the SFTP server. A full example would be:

**Server**
```
$ chisel server --sftp
```

**Client**
```
$ cat chisel-sftp.sh
#!/bin/sh
nc 127.0.0.1 2222

$ chisel client http://server sftp

$ sftp -S ./chisel-sftp.sh bogus
sftp > 
```

The `bogus` parameter is not used, but required.

I realize this is all sort of hacky, but I couldn't think of any other way to make SFTP over chisel work without double SSH encapsulation. I also just considered a FTP endpoint instead, but that gets hairy since it uses multiple ports. Hopefully this SFTP feature, it is useful to someone else. 

All of this also made me sort of realize that with port forwarding, `socks`, `stdio`, and now `sftp`, chisel is becoming much like what `ssh` traditionally provides. Should `pty` be next? 😃 

@jpillora let me know what you think.


jpillora:
Hey @aus this PR looks good, my only concern is the DX of the netcat shell script. If the script is required, then could the script use stdio mode instead?

I'm not very familiar with sftp though -S looks to be equivalent to ssh proxy command?

ssh -o ProxyCommand='chisel client chisel-server.com stdio:%h:%p' user@localhost

BoleynSu:
I personally would like to avoid a built-in sftp due to the complexity introduced. The more complex the software is, the more likely it will be buggy. For myself, I use chisel only for websocket tunneling my sshd service. If I want sftp, I would enable it on my sshd server instead.

BoleynSu:
Very good point for the double encapsulation of SSH. I think it is what we want to fix. I stopped using chisel because of it and instead implemented my own websocket tunnel as follows in less than 200 lines. I think the better solution would be using a similar approach in chisel as well. I did not have time to port it to chisel yet. Maybe someone who is interested can do this or I will do it later (no guarantees). 

```golang
package main

import (
	"errors"
	"flag"
	"io"
	"log"
	"net"
	"net/http"
	"net/url"
	"os"
	"sync"
	"time"

	"github.com/gorilla/websocket"
)

var remote = flag.String("remote", "localhost:22", "remote address")
var local = flag.String("local", "stdio", "local address")
var addr = flag.String("addr", "localhost:8080", "service address")
var wss = flag.Bool("wss", false, "use wss")
var srv = flag.Bool("srv", false, "run as server")

type WSConn struct {
	*websocket.Conn
	buf []byte
}

func NewWSConn(wsConn *websocket.Conn) *WSConn {
	return &WSConn{Conn: wsConn}
}
func (c *WSConn) Read(p []byte) (int, error) {
	if c.buf == nil {
		mt, msg, err := c.ReadMessage()
		if err != nil {
			return 0, err
		}
		if mt != websocket.BinaryMessage {
			return 0, errors.New("message type mismatch")
		}
		c.buf = msg
	}
	n := copy(p, c.buf)
	if n == len(c.buf) {
		c.buf = nil
	} else {
		c.buf = c.buf[n:]
	}
	return n, nil
}
func (c *WSConn) Write(p []byte) (int, error) {
	if err := c.WriteMessage(websocket.BinaryMessage, p); err != nil {
		return 0, err
	}
	return len(p), nil
}

type CloseFunc func() error

func (cf CloseFunc) Close() error {
	return cf()
}

func Pipe(src io.ReadWriteCloser, dst io.ReadWriteCloser) {
	var wg sync.WaitGroup
	var o sync.Once
	close := func() {
		src.Close()
		dst.Close()
	}
	wg.Add(2)
	go func() {
		io.Copy(src, dst)
		o.Do(close)
		wg.Done()
	}()
	go func() {
		io.Copy(dst, src)
		o.Do(close)
		wg.Done()
	}()
	wg.Wait()
}

func Server() {
	http.HandleFunc("/",
		func(w http.ResponseWriter, r *http.Request) {
			var upgrader = websocket.Upgrader{}
			c, err := upgrader.Upgrade(w, r, nil)
			if err != nil {
				log.Print("Upgrade:", err)
				return
			}
			conn, err := net.Dial("tcp", *remote)
			if err != nil {
				c.Close()
				log.Print("Dial:", err)
				return
			}
			Pipe(conn, NewWSConn(c))
		})
	log.Fatal("ListenAndServe:", http.ListenAndServe(*addr, nil))
}

func Client() {
	handle := func(conn io.ReadWriteCloser) {
		wait := 10 * time.Millisecond
		for i := 0; i < 15; i++ {
			u := url.URL{Scheme: "ws", Host: *addr, Path: "/"}
			if *wss {
				u.Scheme = "wss"
			}
			c, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
			if err != nil {
				log.Print("Dial:", err)
				log.Print("wait for ", wait)
				time.Sleep(wait)
				wait *= 2
			} else {
				Pipe(conn, NewWSConn(c))
				return
			}
		}
		conn.Close()
	}
	if *local == "stdio" {
		handle(&struct {
			io.Reader
			io.Writer
			io.Closer
		}{
			os.Stdin,
			os.Stdout,
			CloseFunc(func() error {
				os.Stdin.Close()
				os.Stdout.Close()
				os.Exit(0)
				return nil
			}),
		})
	} else {
		l, err := net.Listen("tcp", *local)
		if err != nil {
			log.Fatal("Listen:", err)
		}
		for {
			conn, err := l.Accept()
			if err != nil {
				log.Print("Accept:", err)
				continue
			}
			go handle(conn)
		}
	}
}

func main() {
	flag.Parse()
	log.SetFlags(0)
	if *srv {
		Server()
	} else {
		Client()
	}
}
```

jpillora:
RE: double ssh: I agree it’ suboptimal, though are we able to get the
throughout cost of the extra ssh layer?

Benchmark with netcat vs plain chisel?

BoleynSu:
Having a benchmark would be great.

Double ssh has its use case where we want to tunnel unencrypted data and I think this is the reason why @jpillora introduced it in the first place. Double ssh issue is more like a multiple encryption issue (another layer of encryption if we also use TLS).

jpillora:
My guess is that double/triple/quadruple encryption shouldn't have a
noticeable impact on performance for most people, especially since (a)
modern processors have built in operations for symmetric crypto and (b)
consuming an extra 10-20 bytes from the average packet (1500 MTU) shouldn't
affect the throughput at low speeds (sub-150Mbit). That said, I'd be
happy for this guess to be proved wrong!

aus:
re: the netcat requirement, you could also do use `stdio` (didn't even think to do this, good suggestion). You'll still need to use netcat with `R:sftp` since `stdio` doesn't work in reverse though.

```
$ cat sftp-chisel.sh
#!/bin/bash
exec go run main.go client http://localhost:8000 stdio:sftp

$ sftp -S ./sftp-chisel.sh bogus
Connected to bogus.
sftp>
```

re: benchmarks, I'll try to throw some together later, but would also expect not much of an impact. While the double encapsulation isn't ideal, I mainly built this to solve a problem of not having access to an sshd. Example: Server needs to download client files over SFTP but the client and/or server does not have ssh daemon. `chisel client http://server r:sftp` solves this. 

re: complexity, totally understand if we want to keep the complexity low, so no hard feelings. But I personally feel chisel has become less TCP-over-WebSockets and more SSH-Channels-over-WebSockets, where SSH channels are arbitrary transports for anything: tcp, udp, socks, sftp, and stdio. Much like what's in [RFC 4254](https://tools.ietf.org/html/rfc4254).

BoleynSu:
I did a very rough benchmark.

```bash
scp -o ProxyCommand='./chisel client localhost:8080 stdio:localhost:22'  /home/boleyn.su/Downloads/Camera\ Roll.zip  
scp -o ProxyCommand='./wstunnel' /home/boleyn.su/Downloads/Camera\ Roll.zip vm.boleyn.su:
```


chisel reaches around 60MB/s while my implementation above reaches around 130MB/s.

BoleynSu:
re @aus, I got your point about not having a sshd daemon. I am totally fine with your implementation except for the complexity introduced. I guess it would be @jpillora's call to make the decision.

> Another approach I considered was enhancing the `stdio` remote to execute a local command. Something like `stdio:/usr/lib/openssh/sftp-server` as the remote. Your `sftp -S` script would then look something like `chisel client http://server:8080 stdio:/usr/lib/openssh/sftp-server`. The downsides of this is that it taints the remote syntax, it gets weird when reversing, and you don't get SFTP on Windows. Might be worth implementing somehow though. Thoughts @BoleynSu?

This approach has security concerns we want to avoid (remote code execution). Thus, if we will add sftp support. I would prefer your current implementation.

jpillora:
> chisel has become less TCP-over-WebSockets and more SSH-Channels-over-WebSockets, where SSH channels are arbitrary transports for anything: tcp, udp, socks, sftp, and stdio

I do agree, and if the 3rd party packages are reliable and maintained I don't mind including them. I'm mostly interested in the developer experience, and hoping for an easier solution than the nc shell script...

Separate question, how does `scp` compare to `sftp`? I mostly use `scp` to transfer files

```
scp -o ProxyCommand='chisel client chisel-server.com stdio:%h:%p' user@localhost:~/my-remote-file my-local-file
```

aus:
My understanding is this: `scp` and `sftp` are different protocols. When you scp a file, it initiates the local scp process, establishes an SSH connection to the remote, then starts a scp process on the remote server. The local and remote scp processes then send and receive data over their established pipes. It's more or less `cat file > remotefile` over SSH.

`sftp` is a little different in that it supports more file operations that make it more like a remote file system protocol. Clients can support resuming interrupted transfers, directory listings, remote file removal, and even mounting. It's also more cross-platform. SFTP is most often implemented as a subsystem to SSH.

But both protocols wrap SSH for their connection. Neither SFTP nor SCP provides authentication and security, because it expects the underlying protocol (most often SSH) to secure this. Both `scp` and `sftp` clients subsequently support the `ProxyCommand` SSH option (among other SSH options). So you can also do a SFTP in this strategy:

```
sftp -oProxyCommand='chisel client chisel-server.com stdio:%h:%p' user@localhost:~/my-remote-file my-local-file
```

However, using the ProxyCommand is going to likely result in double encapsulation of SSH, which has some performance hits (though possibly acceptable).

I agree that connecting to the `sftp` remote is a bit weird. The expectation from users is probably just to point their sftp client to the listening port and be on their way.  But the only way I see that working is with double SSH and initiating the remote's SFTP server (if it exists).

The benefits of having a `sftp` remote boil down to:

- No double SSH
- Built in SFTP server (where there otherwise may be none, ie Windows)

And due to the way the sftp(1) client works, you have to use the -S option like this:

```
$ cat sftp-chisel.sh
#!/bin/bash
exec chisel client chisel-server.com stdio:sftp

$ sftp -S ./sftp-chisel.sh bogus
Connected to bogus.
sftp>
```

Or if you need the sftp remote in reverse mode, you have to go netcat (or likewise):

_Client_
```
$ chisel client chisel-server.com R:sftp
```
_Server_
```
$ cat chisel-sftp.sh
#!/bin/sh
nc 127.0.0.1 2222

$ sftp -S ./chisel-sftp.sh bogus
Connected to bogus.
sftp> 
```

Issue #197, opened 2020-09-18
IPv6 remotes?


shawnhwei:
Seems like the chisel client doesn't like ipv6 remote addresses (e.g. [::1]:8888).

Based on [here](https://github.com/jpillora/chisel/blob/6fce8e1e6e3bf2da6f32eaaaab692022137319d0/share/settings/remote.go#L49) I'm guessing this is to be expected.

Or maybe I'm doing it wrong?

jpillora:
True, v6 should be supported, so this is a bug. PR? Otherwise I’ll work on
it when I get a chance next

PR #198, opened 2020-09-19
IPv6 remotes support

shawnhwei:
I don't touch Go much but it seems like this maybe all that's needed enable IPv6 remotes (#197).

### DIFF
--- a/share/settings/remote.go
+++ b/share/settings/remote.go
@@ -46,7 +46,7 @@ func DecodeRemote(s string) (*Remote, error) {
 		s = strings.TrimPrefix(s, revPrefix)
 		reverse = true
 	}
-	parts := regexp.MustCompile(`(\[[A-Za-z0-9:/]+\]|[^\s:]+):?`).FindAllStringSubmatch(s, -1)
+	parts := regexp.MustCompile(`(\[[^\[\]]+\]|[^\[\]:]+):?`).FindAllStringSubmatch(s, -1)
 	if len(parts) <= 0 || len(parts) >= 5 {
 		return nil, errors.New("Invalid remote")
 	}
--- a/share/settings/remote.go
+++ b/share/settings/remote.go
@@ -85,7 +85,7 @@ func DecodeRemote(s string) (*Remote, error) {
 			return nil, errors.New("Missing ports")
 		}
 		if !isHost(p) {
-			return nil, errors.New("Invalid host")
+			return nil, errors.New("Invalid host" + p)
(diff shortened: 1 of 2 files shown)


share/settings/remote.go
```diff
@@ -85,7 +85,7 @@ func DecodeRemote(s string) (*Remote, error) {
 			return nil, errors.New("Missing ports")
 		}
 		if !isHost(p) {
-			return nil, errors.New("Invalid host")
+			return nil, errors.New("Invalid host" + p)
```
jpillora, line 88: this could return `fmt.Errorf("Invalid host (%s)", p)`

share/settings/remote.go
```diff
@@ -46,7 +46,7 @@ func DecodeRemote(s string) (*Remote, error) {
 		s = strings.TrimPrefix(s, revPrefix)
 		reverse = true
 	}
-	parts := strings.Split(s, ":")
+	parts := regexp.MustCompile(`([A-Za-z0-9./]+|\[[A-Za-z0-9:/]+\]):?`).FindAllStringSubmatch(s, -1)
```
jpillora, line 49: Is `[A-Za-z0-9./]+` all valid characters for a hostname?

Maybe put the `[hex and colons]` match on the left hand side of the OR `|` and then make the hostname matcher more lax?

share/settings/remote_test.go
```diff
@@ -81,6 +81,16 @@ func TestRemoteDecode(t *testing.T) {
 			},
 			"localhost:5353:1.1.1.1:53/udp",
 		},
+		{
+			"[::1]:8080:google.com:80",
+			Remote{
+				LocalHost:  "[::1]",
+				LocalPort:  "8080",
+				RemoteHost: "google.com",
```
jpillora, line 89: I'm sure it'd work, though could you please add another test with v6 in both local and remote hosts?

share/settings/remote.go
shawnhwei: My bad, forgot to remove this after doing some testing.

share/settings/remote.go
shawnhwei: That's a good point, this misses at least UTF8 domains and domains with hyphens.

share/settings/remote_test.go
shawnhwei: Done.

jpillora:
Thanks! Merged, releasing as `v1.7.1`

Issue #149, opened 2020-04-09
chisel do not support ntlm proxy authentication


pariseed:
is in roadmap the possibility to introduce support for ntlm proxy authentication ?

pariseed:
closing this issue because i solve the ntlm proxy auth with:
> https://github.com/launchdarkly/go-ntlm-proxy-auth.git

i made a fork of chisel where i've implemented the above lib in chisel code.
if someone else will have same necessity in future, keep an eye on my fork:

> https://github.com/pariseed/chisel.git

mpgn:
@pariseed why not a PR ?

pariseed:
@mpgn when i've opened this issue i've wait almost a month for an answer from @jpillora but while for other issue opened after that a response was provided for this issue nothing is come.
So i decided to do it from myself, and i've intend the no response as an "i'm not interessed in that feature".
If jpillora let me know he is interessed in that feature i will pleasure to send a PR to him, but until then if someone need ntml integration they can look at my fork.

jpillora:
Hey @pariseed I'd consider a PR, though I'd want to confirm the API beforehand, and also that we don't introduce many new dependencies

pariseed:
Hi @jpillora i understood, i think that if you're sure to use the API there is no sense to forward a PR.
However you can see here the lines that i have added to the client in order to make ntlm work with chisel:

[include ntlm lib ](https://github.com/pariseed/chisel/blob/0b603d8f64977f4d1d3a3d190b4a88b801c9349d/client/client.go#L17)

[add new regex syntax to recognize ntlm invocation](https://github.com/pariseed/chisel/blob/0b603d8f64977f4d1d3a3d190b4a88b801c9349d/client/client.go#L99-L123)

[adding the ntlmDialContext](https://github.com/pariseed/chisel/blob/0b603d8f64977f4d1d3a3d190b4a88b801c9349d/client/client.go#L253-L268)


If you change idea let me know and i will create a patch from this lines or a PR

jpillora:
A few things changes that I'd request:

* github.com/launchdarkly/go-ntlm-proxy-auth uses https://github.com/launchdarkly/go-ntlmssp - it would be better to depend directly on Microsoft's module https://github.com/Azure/go-ntlmssp, and maybe add a small wrapper in chisel if needed

* the user interface is essentially this regex `(NTLM)þ(.*):(.*):(.*)@`, since there is no `ntlm:` protocol, we could simply just use that. so it would be better to:
[...]

PR #204, opened 2020-10-20
Ntlm proxy auth

pariseed:
following almost all the suggestion argued in [issue 149](https://github.com/jpillora/chisel/issues/149) this pull request enable chisel to connect on ntlm v2 proxies

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -263,14 +263,14 @@ func (c *Client) setProxy(u *url.URL, d *websocket.Dialer) error {
 	// CONNECT proxy
 	if !strings.HasPrefix(u.Scheme, "socks") {
 		//check if ntlm connection is required
-		if c.Isntlm {
-			dialertime := &net.Dialer{
-			    Timeout:   30 * time.Second,
-			    KeepAlive: 30 * time.Second,
-			}
+                if c.config.Isntlm {
+                        dialertime := &net.Dialer{
+                            Timeout:   30 * time.Second,
+                            KeepAlive: 30 * time.Second,
+                        }
 
-			d.NetDialContext = ntlm.WrapDialContext(dialertime.DialContext, c.config.ntlmurl, c.config.ntlmusr, c.config.ntlmpwd, c.config.ntlmdomain)
-		}
+                        d.NetDialContext = ntlm.WrapDialContext(dialertime.DialContext, c.config.Ntlmurl, c.config.Ntlmusr, c.config.Ntlmpwd, c.config.Ntlmdomain)
+                }
 
 		d.Proxy = func(*http.Request) (*url.URL, error) {
 			return u, nil
--- a/client/client.go
+++ b/client/client.go
@@ -31,13 +31,6 @@ import (
 
 
 
-// ntlm global variables 
-var isntlm bool
-var ntlmdomain string
-var ntlmusr string
-var ntlmpwd string
-var ntlmurl string
-
 
 //Config represents a client configuration
 type Config struct {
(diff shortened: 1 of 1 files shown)


jpillora:
hey @pariseed thanks for the PR, can you run `go mod tidy` in the project root and commit the changed go mod/sum?

pariseed:
Hi @jpillora done, let me know if it's ok :)

jpillora:
Thanks! hopefully can merge soon

jpillora:
Hey guys I remember looking into this and was going to make a few changes myself (though ran out of time):

1. use https://github.com/Azure/go-ntlmssp as a dependency rather than an unknown fork
1. put changes required for 1. into a separate ntlm file
1. move variables into an ntlm configuration struct 
1. fix regex

@pariseed feel free to make these changes if you're up for it

pariseed:
@jpillora as explained in the [issue 149](https://github.com/jpillora/chisel/issues/149) i was not able to successfully use https://github.com/Azure/go-ntlmssp in order to correctly authenticate on ntlm proxys and in the issue i've already explained why this happen, beyond that the guys from [launchdarkly](https://github.com/launchdarkly/go-ntlmssp) didn't do anything of strange on their fork adding only the negotiations flag needed to let authentication work and if they in the future for any reason will messed up their things will be very easily for us untie from them.
Putting all ntlm stuff with the required changes in one separated file i think is not a good idea because it will take a lot work and at the end will lead us to create an our fork without any kind of benefit and making difficult update they if on the main repo other commit will be added. 
Regarding the regex interface instead, i used exactly the one that you've requested on  [issue 149 regex comment](https://github.com/jpillora/chisel/issues/149#issuecomment-712433020) so why after so long time you're requesting change on they ? And what kind of fix do you need ?  
After one year i think that some more explanations is due.
Do not take me wrong but i've already spend a lot of time working on that, i understand that you want to do things in your way and as better they can, but since day1 i see that you've zero interest in this implementation and the time passed by i think that prove that, so i will not put any further effort on this implementation.
Hope you can figure out.

jpillora:
1. "i used exactly the one that you've requested onissue 149 regex comment "

`htt.*://` no, this is wrong

2.

```
if len(ntlmfind) == 0 {
         ;;
 } else {
```

why?

3.

```
                        c.Isntlm = true
                         c.Ntlmdomain = ntlmfind[1]
                         c.Ntlmusr = ntlmfind[2]
                         c.Ntlmpwd = ntlmfind[3]
```
variable casing is wrong
why are they public? 

---

oct 2020 i said that i want to make sure "we don't introduce many new dependencies"

id be okay accepting a single dependency from Microsoft (Azure) - this means we **copy** the changes that `launchdarkly` introduced (["A little copying is better than a little dependency."](https://go-proverbs.github.io)). i said this back then because 1. we get official fixes from microsoft - we dont rely on launchdarkly, and 2. the changes are minimal: its essentially this [file](https://github.com/launchdarkly/go-ntlm-proxy-auth/blob/90effd9/ntlm.go), where the default flags have [been changed](https://github.com/Azure/go-ntlmssp/compare/master...launchdarkly:master#diff-a03ba36fa5d352de04edb66e4d2c30d5bacffd3a39456653d5b0edea79b43588). so instead of using his fork where he changes 2 lines to get different defaults in order to change the internal behaviour of `ntlmssp.NewNegotiateMessage` - you just construct your own a `NegotiateMessage` yourself.

in summary, the PR should include be the changes you've got right now, plus the fixes, plus this 100 line file from launchdarkly/go-ntlm-proxy-auth with the flag changes so we can use the original azure repo.

i do open source in my free time, and time is very valuable, so i need contributors to put in the work to enhance chisel - while keeping it stable, without introducing bad dependencies, bugs, and bad code

PR #207, opened 2020-10-26
Support Unix Domain Socket forwarding

nguyentruongtho:
Forwarding with Unix Domain Socket instead of TCP can improve the performance in some cases (eg: [mysql](https://dev.mysql.com/doc/refman/5.6/en/can-not-connect-to-server.html#:~:text=A%20Unix%20socket%20file%20connection,can%20connect%20using%20TCP%2FIP.))

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,21 @@ $ chisel client --help
     R:<local-interface>:<local-port>:<remote-host>:<remote-port>
 
   which does reverse port forwarding, sharing <remote-host>:<remote-port>
-  from the client to the server's <local-interface>:<local-port>.
+  from the client to the server's <local-interface>:<local-port>, or:
+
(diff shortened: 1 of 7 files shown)


huapox:
+1

huapox:
> @jpillora this seems like a reasonable PR. Any chance you'd like to give it a review?

Current reuse socket like `/tmp/xx1.socket` in RemotePort field, which make the code hard-reading, may be a refactor of the struect `Remote` .

Or complete-define the format of connect string:

- current: `R:127.0.0.1:xxx:unix:///tmp/xx1.sock`
- maybe? `R:bindURI:remoteURI` both bindURI/remoteURI can be TCP/UNIX mode:
  - tcpMode: xxxhost:xxxport
  - unixMode: unix:///tmp/xx1.socket

PR #215, opened 2020-11-09
Use SHA256 hashes for key fingerprints

### LINKED ISSUE #214: Use SHA256 Hash for Key Fingerprints
How would you feel about using the `SHA256` hash instead of the current `MD5` to calculate public key fingerprints?
https://github.com/jpillora/chisel/blob/v1.7.2/share/ccrypto/keys.go#L35

To remain backward compatible we could add a `sha256:` prefix to new fingerprints and evaluate old fingerprints (without the prefix) still as md5 hashes (at least until a new major version would allow the breaking change).

References regarding MD5 deprecation:
* https://tools.ietf.org/id/draft-lvelvindron-tls-md5-sha1-deprecate-01.html

srueg:
Closes #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,9 +244,8 @@ $ chisel client --help
 
   Options:
 
-    --fingerprint, A *strongly recommended* fingerprint string
-    to perform host-key validation against the server's public key.
-    You may provide just a prefix of the key or the entire string.
+    --fingerprint, A *strongly recommended* fingerprint (base64 encoded SHA256)
+    string to perform host-key validation against the server's public key.
     Fingerprint mismatches will close the connection.
 
     --auth, An optional username and password (client authentication)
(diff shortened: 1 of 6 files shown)


jpillora:
Thanks for the PR though this breaks existing clients using `--fingerprint`

jpillora:
Verification should look for hash outputs like this https://play.golang.org/p/7dSkGjdSNO-

1. 44 chars, ends with `=`, and then base64 decode, verify sha256
1. otherwise, hex decode, verify md5

jpillora:
Also we should update CLI help and docs (mention sha256, undocument md5 hex examples, and support for hash prefixes)

srueg:
Yes, absolutely. Sorry, I wanted to mark the PR as a draft until all the things you mention get implemented.

jpillora:
No worries 👍

jpillora:
PR looks good! Will test soon

jpillora:
Thanks :) Merged and released in `v1.7.3`

Issue #75, opened 2019-01-08
Disable Encryption


0xjohn:
I want to disable encryption and accept connection without ssh handshake, is it possible to disable chisel encryption?

matti:
Isn't chisel connection secure when the server runs behind https (or wss) - disabling encryption would make sense in that situation.

jpillora:
Yep, it would make sense to disable encryption if you're already running over HTTPS. Unfortunately, SSH [appears not to support a NULL cipher](https://github.com/golang/crypto/blob/4def268fd1a49955bfb3dda92fe3db4f924f2285/ssh/common.go#L28-L42) so it would be a very large change to use an alternate multiplex/control protocol, most likely a rewrite.

PR #226, opened 2020-12-10
ssh: allow to configure ciphers/macs

JulienBalestra:
This PR allows to reduce the footprint of **chisel** when using **arm** devices without any `aes*` hardware support.

It allows to reduce the ssh cpu cipher footprint to `arcfour` and tune the macs accordingly.
This use case is motived by an isolated location using a 4G connection with a low power setup. Many 4G setups aren't directly exposed to the internet, making the usual NAT impossible.

This is going to increase the throughput of the setup. Especially when the remote application already has an encrypted setup, the additional encryption added by ssh isn't needed anymore as mentioned in https://github.com/jpillora/chisel/issues/75.

The `none` encryption cipher requires more work as it's not supported by the underlying used ssh library.

I'll send some measurements with `iperf` anytime soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,12 @@ $ chisel server --help
     authfile with {"<user:pass>": [""]}. If unset, it will use the
     environment variable AUTH.
 
+    --ciphers, An optional string comma separated to provide a list of ciphers to
+    use by order of preference and matching the client (default "aes128-gcm@openssh.com,chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr").
+
+    --macs, An optional string comma separated to provide a list of MACs to
+    use by order of preference and matching the client (default "hmac-sha2-256-etm@openssh.com,hmac-sha2-256,hmac-sha1,hmac-sha1-96").
+
     --keepalive, An optional keepalive interval. Since the underlying
     transport is HTTP, in many instances we'll be traversing through
     proxies, often these proxies will close idle connections. You must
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -384,7 +225,200 @@ var clientHelp = `
...
+
+    --help, This help text
+
+  Signals:
+    The chisel process is listening for:
+      a SIGUSR2 to print process stats, and
+      a SIGHUP to short-circuit the client reconnect timer
+
+  Version:
+    ` + chshare.BuildVersion + ` (` + runtime.Version() + `)
```
jpillora, line 292: Note: Instead of using string placeholders, we just concat with function calls

main.go
```diff
@@ -109,6 +42,12 @@ var serverHelp = `
     of man-in-the-middle attacks (defaults to the CHISEL_KEY environment
     variable, otherwise a new key is generate each run).
 
+    --ciphers, An optional string comma separated to provide a list of ciphers to
+    use by order of preference and matching the client (default %q).
```
jpillora, line 46: Could you please switch the `%q` to ` + defaultCiphers() + `,
similarly with `defaultMACs()`
then we don't need the `fmt.Sprintf` - I think doesn't scale well with large strings, as you'll struggle to find all the placeholders then match them up with the ordered arguments.

I think the more scalable solution would be Go text templates. Then we can use named placeholders `{{ .DefaultMACs }}` etc. However for this, I think the string concat is the simplest and most readable.

README.md
```diff
@@ -258,6 +264,12 @@ $ chisel client --help
     the credentials inside the server's --authfile. defaults to the
     AUTH environment variable.
 
+    --ciphers, An optional string comma separated to provide a list of ciphers to
```
jpillora, line 267: Question: Is it common to customise with  1 or 2 ciphers? or is it more common to provide them all, and maybe change the order?

I ask because an alternate CLI would be `--cipher` where you can it provide multiple times. Each flag appends the `[]Ciphers` list. Though I'd only recommend this if the common case was to specify 1 or 2.

JulienBalestra: In my opinion, you provide a list where the order matter, you could have a heterogeneous fleet of remote supporting different ciphers but you want them to be taken into account with the server.

main.go
JulienBalestra: Agreed, I'm ready to make the change for the go template one, seems more scalable to me too, sound good to you ?

README.md
jpillora: Okay no worries, a CSV string works 👍

main.go
jpillora: Sorry for the delay! Fire away with the Go template solution

PR #231, opened 2020-12-28
Upload a sha256sum.txt file for every new release

znerol:
Docs here: https://goreleaser.com/customization/checksum/


znerol:
Nevermind, checksums are already there :/

Issue #232, opened 2020-12-30
bind failure results in panic


matti:
```
/app # socat TCP4-LISTEN:1080 -
2020/12/30 22:02:31 socat[150] E bind(5, {AF=2 0.0.0.0:1080}, 16): Address in use
```

```
/app # /app/chisel server --port 1080
2020/12/30 21:59:35 server: Fingerprint DYMxPJ33dLHCDEUEyEtg2qd7Pq+Jmz/XxSj/V7/yWjs=
panic: runtime error: invalid memory address or nil pointer dereference
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x70c36d]

goroutine 8 [running]:
net/http.(*onceCloseListener).close(...)
        /usr/local/go/src/net/http/server.go:3371
sync.(*Once).doSlow(0xc000124010, 0xc00006abf0)
        /usr/local/go/src/sync/once.go:66 +0xec
sync.(*Once).Do(...)
        /usr/local/go/src/sync/once.go:57
net/http.(*onceCloseListener).Close(0xc000124000, 0x0, 0x0)
        /usr/local/go/src/net/http/server.go:3367 +0x78
panic(0x8abfc0, 0xc41880)
        /usr/local/go/src/runtime/panic.go:969 +0x1b9
net/http.(*onceCloseListener).Accept(0xc000124000, 0xc000026038, 0x8a1b20, 0xc417c0, 0x91a200)
        <autogenerated>:1 +0x32
net/http.(*Server).Serve(0xc000020380, 0x0, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/http/server.go:2937 +0x266
github.com/jpillora/chisel/share/cnet.(*HTTPServer).GoServe.func1(0x0, 0x0)
        /src/share/cnet/http_server.go:54 +0x3f
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc0000ee570, 0xc00000d8a0)
        /go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
        /go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup.go:54 +0x66
```

```
/app # /app/chisel server --port 1081
[...]

PR #234, opened 2021-01-01
Fail gracefully when attempting to start server on an address/port already in use

oh6hay:


oh6hay:
hi @jpillora have you had a chance to look at this?

server/server.go
```diff
@@ -142,6 +142,7 @@ func (s *Server) StartContext(ctx context.Context, host, port string) error {
 	}
 	l, err := s.listener(host, port)
 	if err != nil {
+		s.Errorf("Cannot listen: %w", err)
```
jpillora, line 145: how does this cause it fail gracefully? this just prints a message doesnt it...

oh6hay: It seems when I originally submitted this PR, my change in server_listen.go fixed an actual issue which you also fixed two weeks later in response to https://github.com/jpillora/chisel/issues/232 so this PR is no longer valid

today I checked the merge conflict on this one, resolved it but didn't realize this was fixed elsewhere

PR #250, opened 2021-03-03
Feature socks udp associate

Meteorite:
Add UDP associate support to built-in socks proxy.

UDP associate protocol (see https://tools.ietf.org/html/rfc1928 , chapters 6 and 7) requires server to check client's UDP packets source IP address and drop packets from unknown IPs (i.e. from IPs, which don't hold active TCP connection to socks server). That's why it is very inconvenient and inefficient to implement this functionality in the current architecture with socks proxy on outgoing party (chisel server for forward socks or chisel client for reverse socks), as it will end up forwarding each packet from incoming party to outgoing via SSH may be only to drop it there. Besides, it would require chisel server to know incoming UDP port numbers of every connected chisel client to send proper replies to associate requests for forward socks.

Therefore, this PR moves socks server from outgoing party to incoming party, leaving only SSH-controlled connector on outgoing party. To do it we need a modular socks server implementation that allows injecting custom request handler and outgoing traffic handler. I failed to find such an implementation as a ready to use library and crafted custom socks server code, based on https://github.com/armon/go-socks5 and its fork with associate support https://github.com/haxii/socks5 . This crafted implementation is located in share/socks5 dir. When you create a socks server, you can pass a custom Handler implementation. Handlers could use one of associate implementation:
- SingleUDPPortAssociate - opens one UDP port, when socks server starts and uses it for the lifetime of server to serve all clients.
- MultiUDPPortAssociate - opens separate UDP port for every client and closes it, when client closes corresponding TCP-session.
For now, SingleUDPPortAssociate is used in PR.

PR does not change any command line arguments to chisel client or chisel server. But one UDP port is opened for every socks remote on client and R:socks remote on server. Free UDP port number is chosen automatically (clients do not need to know it in advance, as they will receive it in replies to UDP associate requests to socks server).

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,7 +1,7 @@
 github.com/andrew-d/go-termutil v0.0.0-20150726205930-009166a695a2 h1:axBiC50cNZOs7ygH5BgQp4N+aYrZ2DNpWZ1KG3VOSOM=
 github.com/andrew-d/go-termutil v0.0.0-20150726205930-009166a695a2/go.mod h1:jnzFpU88PccN/tPPhCpnNU8mZphvKxYM9lLNkd8e+os=
-github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 h1:0CwZNZbxp69SHPdPJAN/hZIm0C4OItdklCFmMRWYpio=
-github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5/go.mod h1:wHh0iHkYZB8zMSxRWpUBQtwG5a7fFgvEO+odwuTv2gs=
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/fsnotify/fsnotify v1.4.9 h1:hsms1Qyu0jgnwNXIxa+/V/PDsU6CfLf6CNO8H7IWoS4=
 github.com/fsnotify/fsnotify v1.4.9/go.mod h1:znqG4EE+3YCdAaPaxE2ZRY/06pZUdp0tY4IgpuI1SZQ=
 github.com/gorilla/websocket v1.4.2 h1:+/TMaTYc4QFitKJxsQ7Yye35DkWvkdLcvGKqM+x0Ufc=
@@ -14,6 +14,12 @@ github.com/jpillora/requestlog v1.0.0 h1:bg++eJ74T7DYL3DlIpiwknrtfdUA9oP/M4fL+Pp
 github.com/jpillora/requestlog v1.0.0/go.mod h1:HTWQb7QfDc2jtHnWe2XEIEeJB7gJPnVdpNn52HXPvy8=
 github.com/jpillora/sizestr v1.0.0 h1:4tr0FLxs1Mtq3TnsLDV+GYUWG7Q26a6s+tV5Zfw2ygw=
 github.com/jpillora/sizestr v1.0.0/go.mod h1:bUhLv4ctkknatr6gR42qPxirmd5+ds1u7mzD+MZ33f0=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0 h1:4G4v2dO3VZwixGIRoQ5Lfboy6nUhCyYzaqnIAPPhYs4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
+github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/tomasen/realip v0.0.0-20180522021738-f0c99a92ddce h1:fb190+cK2Xz/dvi9Hv8eCYJYvIGUTN2/KLq1pT6CjEc=
 github.com/tomasen/realip v0.0.0-20180522021738-f0c99a92ddce/go.mod h1:o8v6yHRoik09Xen7gje4m9ERNah1d1PPsVq1VEx9vE4=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
@@ -33,3 +39,6 @@ golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae h1:Ih9Yo4hSPImZOpfGuA4bR/ORK
 golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/text v0.3.0 h1:g61tztE5qeGQ89tm6NTjjM9VPIm088od1l6aSorWRWg=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c h1:dUUwHk2QECo/6vqA44rthZ8ie2QXMNeKRTHCNY2nXvo=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
--- a/share/tunnel/socks_handler.go
+++ b/share/tunnel/socks_handler.go
@@ -1,0 +1,135 @@
+package tunnel
+
+import (
+	"context"
+	"encoding/gob"
+	"fmt"
+	"github.com/jpillora/chisel/share/cio"
+	"github.com/jpillora/chisel/share/socks5"
+	"github.com/jpillora/chisel/share/socks5/scope"
+	"golang.org/x/crypto/ssh"
+	"log"
+	"net"
+)
+
+type chiselSocksHandler struct {
+	p            *Proxy
+	udpLocalAddr *socks5.AddrSpec
+	sl           *log.Logger
+	udp          *socks5.SingleUDPPortAssociate
+}
+
+func makeChiselSocksHandler(p *Proxy, localUDPAddr *net.UDPAddr, sl *log.Logger) *chiselSocksHandler {
+	return &chiselSocksHandler{
+		p: p,
+		udpLocalAddr: &socks5.AddrSpec{
+			IP:   localUDPAddr.IP,
+			Port: localUDPAddr.Port,
+		},
+		sl: sl,
+	}
+}
+
+func (h *chiselSocksHandler) OnStartServe(ctxServer socks5.ContextGo, _ net.Listener) error {
+	h.udp = socks5.MakeSingleUDPPortAssociate(h.udpLocalAddr, h, h.sl)
+	return h.udp.ListenAndServeUDPPort(ctxServer, "udp")
+}
+
+func (h *chiselSocksHandler) ErrLog() socks5.ErrorLogger {
+	return h.sl
+}
+
+func (h *chiselSocksHandler) OnConnect(ctx context.Context, conn net.Conn, req *socks5.Request) error {
+	return h.p.pipeRemote(ctx, conn, req.DestAddr.Address()+"/sot", func(dst ssh.Channel) error {
+		code := []byte{0}
+		_, err := dst.Read(code)
+		if err != nil {
+			return fmt.Errorf("can't receive socks code from server: %w", err)
+		}
+		if code[0] != socks5.ReplySucceeded {
+			if err = req.SendError(conn, code[0]); err != nil {
+				return fmt.Errorf("failed to send reply to client: %w", err)
+			}
+			return fmt.Errorf("can't connect to destination server (code: %d)", code[0])
+		}
+		if err := req.SendConnectSuccess(conn); err != nil {
+			return fmt.Errorf("failed to send reply to client: %w", err)
+		}
+		return nil
+	})
+}
+
+func (h *chiselSocksHandler) OnAssociate(_ context.Context, conn net.Conn, _ *socks5.Request) error {
+	return h.udp.OnAssociate(conn)
+}
+
+func (h *chiselSocksHandler) MaxUDPPacketSize() uint {
+	return maxMTU
+}
+
+
+type socksUdpConnector struct {
+	*cio.Logger
+	outbound *udpChannel
+}
+
+func (h *chiselSocksHandler) MakeRemoteUDPConn(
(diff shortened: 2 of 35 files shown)


aus:
This looks awesome, especially considering the amount of changes required to pull this off. I will do some testing with it and report back.

jpillora:
👍

hdid:
I tested it and works well. Please merge this asap 👍
[chisel-socks-udp-1.7.6-linux_amd64.gz](https://github.com/jpillora/chisel/files/6560565/chisel-socks-udp-1.7.6-linux_amd64.gz)

bitsadmin:
Great addition, @Meteorite, that was exactly what I was looking for and it works like a charm! 🥇 @jpillora, can you merge this to the main Chisel repository? :-)

jpillora:
hey @aus did you get a chance to look at this PR?

my main concern is that we are now inheriting the complexity of those SOCKS modules, id prefer to first PR the required features into an existing maintained SOCKS module, and then we depend on that. @Meteorite  i cant see the diff to the SOCKS modules - are there many changes?

aus:
I did some testing a while back and I found no issues with the new SOCKS5 UDP ASSOCIATE feature.

Valid point on inheriting the complexity. To level set, chisel currently uses armon/go-socks5@e75332964ef517daa070d7c38a9466a0d687e0a5 for its SOCKS implementation. A lot of the changes in this PR refect the original code in armon/go-socks5. @Meteorite based his code off of [haxii/socks5](https://github.com/haxii/socks5) which added the UDP ASSOCIATE support. Here is a diff between the two:

https://github.com/armon/go-socks5/compare/master...haxii:master

However, having attempted this implementation myself, these libraries don't work out of the box with chisel. There are some hoops to jump with typing and not having access to some private functions.

So I guess our options are:

a) forking haxii/socks5 into its own module, with necessary changes to support calling from chisel
b) bring in the socks5 code into chisel codebase (as done here in this PR)
c) PR haxii/socks5 to make integration easier

I guess the advantages of option A are that we reduce complexity of chisel's codebase, and we can blame a module if something breaks. 😄 I think option B is worth weighing considering how tightly coupled chisel the remote offerings nowadays.

Meteorite:
share/socks5 dir can be extracted as a library rather easily, if option A is to be taken. I can do it, if needed.

But, indeed, coupling is rather tight, and some features in chisel might require socks5 subsystem modification. So, it might be more convenient to have its code in the same project. At least it was far more convenient during this feature development :)

jpillora:
Thanks for reviewing @aus 

Given that socks is a fairly old protocol - my hope is there won't be changes. So I'm leaning towards option A. @Meteorite I might take you up on that offer, can we move 'share/socks' over to 'github.com/Meteorite/socks'?

Meteorite:
Ок, I'll do it.

go.sum
```diff
@@ -14,6 +14,15 @@ github.com/jpillora/requestlog v1.0.0 h1:bg++eJ74T7DYL3DlIpiwknrtfdUA9oP/M4fL+Pp
...
 github.com/jpillora/sizestr v1.0.0 h1:4tr0FLxs1Mtq3TnsLDV+GYUWG7Q26a6s+tV5Zfw2ygw=
 github.com/jpillora/sizestr v1.0.0/go.mod h1:bUhLv4ctkknatr6gR42qPxirmd5+ds1u7mzD+MZ33f0=
+github.com/meteorite/scope v0.0.0-20210314203727-1e230fea59ae h1:+OY7loRJNAPnyYG4YoPn1wuAvX4SHeYo2XyIQmHJSWM=
+github.com/meteorite/scope v0.0.0-20210314203727-1e230fea59ae/go.mod h1:oGJ2Lf7WjcbEgwaANS+H3ykcLytXXiuHEWLfalSzXb8=
+github.com/meteorite/socks5 v0.0.0-20210604215257-bf325eecbc5d h1:0CEORKRd169DDvpucqMLH57o9UPGMF8Akw3W3YLJ1Kk=
+github.com/meteorite/socks5 v0.0.0-20210604215257-bf325eecbc5d/go.mod h1:1q/GEvRbr06wBgq8tHcvGKIC+AFTZJrwj73HPfF2LwU=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
```
jpillora, line 24: Could you please `go mod tidy`

share/tunnel/socks_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	"fmt"
+	"github.com/jpillora/chisel/share/cio"
+	"github.com/meteorite/scope"
+	"github.com/meteorite/socks5"
+	"golang.org/x/crypto/ssh"
+	"log"
+	"net"
+)
+
+type chiselSocksHandler struct {
```
jpillora, line 15: can we rename to `socksHandler`?

share/tunnel/socks_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	p            *Proxy
+	udpLocalAddr *socks5.AddrSpec
+	sl           *log.Logger
+	udp          *socks5.SingleUDPPortAssociate
+}
+
+func makeChiselSocksHandler(p *Proxy, localUDPAddr *net.UDPAddr, sl *log.Logger) *chiselSocksHandler {
```
jpillora, line 22: can we rename to `newSocksHandler`?

share/tunnel/tunnel_out_ssh.go
```diff
@@ -74,8 +82,19 @@ func (t *Tunnel) handleSSHChannel(ch ssh.NewChannel) {
 	l.Debugf("Close %s%s", t.connStats.String(), errmsg)
 }
 
-func (t *Tunnel) handleSocks(src io.ReadWriteCloser) error {
-	return t.socksServer.ServeConn(cnet.NewRWCConn(src))
+func (t *Tunnel) handleSocksTCP(l *cio.Logger, src io.ReadWriteCloser, hostPort string) error {
```
jpillora, line 85: Interesting, so this function now reimplements `ServeConn` ?

share/tunnel/tunnel_in_proxy.go
```diff
@@ -52,6 +58,21 @@ func (p *Proxy) listen() error {
...
+				_ = l.Close()
+				return p.Errorf("resolve: %s", err)
+			}
+
+			sl := log.New(ioutil.Discard, "", 0)
+			if p.Logger.Debug {
+				sl = log.New(os.Stdout, "[socks]", log.Ldate|log.Ltime)
+			}
+			p.socksServer, _ = socks5.New(&socks5.Config{
+				Handler: makeChiselSocksHandler(p, udpAddr, sl),
```
jpillora, line 73: If this code does belong in Proxy, then Instead of using `sl` we should use `p.Logger.Fork("socks")`

jpillora (commented):
Not enough time to fully review

Could you provide a summary of how SOCKS UDP works? I think I'm missing the forrest for the trees...

jpillora:
Ah just noticed the big summary in the original description sorry

jpillora:
> Free UDP port number is chosen automatically (clients do not need to know it in advance, as they will receive it in replies to UDP associate requests to socks server).

So this means, if I open port `443` on my firewall to UDP, and open remote `443 -> socks`, this won't work? because it will use random ports?

go.sum
Meteorite: Already did it. `testify` is here in `go.sum`, because new dependencies (`github.com/meteorite/scope` and `github.com/meteorite/socks5`) use it for their tests:

```
> go mod graph | grep testify
github.com/meteorite/scope@v0.0.0-20210314203727-1e230fea59ae github.com/stretchr/testify@v1.7.0
github.com/meteorite/socks5@v0.0.0-20210604215257-bf325eecbc5d github.com/stretchr/testify@v1.7.0
...
```

so, it is a transitive dependency. Note, it is not in `go.mod`.

share/tunnel/socks_handler.go
Meteorite: ok

share/tunnel/socks_handler.go
Meteorite: ok

aus:
I will point out though that the “UDP firewall problem” is not specific to chisel or the implementation here. You would have the same problem in a standalone RFC-compliant SOCKS5 server.

aus:
> It seems, we need some cmdline way either to explicitly enable socks UDP (per remote or per all remotes) or to explicitly specify UDP port for socks remote.

doesn’t this need to be a port range rather than a specific UDP port?

it’s my understanding the UDP associate socks command works like this:

1. socks client opens TCP connection to socks server 
2. Client issues UDP ASSOCIATE command requesting destination 
3. Server replies with free UDP port for client to use 
4. Client send UDP packet to socks sever on port from step 3 reply
5. Socks server forward UDP packet to destination


If we use a single port, how would the server keep track of the multiple clients or multiple destinations on a single UDP ASSOCIATE port?

Meteorite:
> it’s my understanding the UDP associate socks command works like this:
> 
>     1. socks client opens TCP connection to socks server
> 
>     2. Client issues UDP ASSOCIATE command requesting destination
> 
>     3. Server replies with free UDP port for client to use
> 
>     4. Client send UDP packet to socks sever on port from step 3 reply
> 
>     5. Socks server forward UDP packet to destination
> 
> 
> If we use a single port, how would the server keep track of the multiple clients or multiple destinations on a single UDP ASSOCIATE port?

You are mostly right, but destination is specified on step 4, not step 2. On step 4 client must sent not an original UDP packet but UDP packet with a special UDP request header (see https://datatracker.ietf.org/doc/html/rfc1928, chapter 7). That header specifies the destination for this particular UDP packet. That is how server can tarck multiple destinations for the same client.

On step 2 some destination might also be specified, but it is only kind of recommendation to server. RFC says:
```
   The server MAY use this information to limit access to the association.  If the
   client is not in possesion of the information at the time of the UDP
   ASSOCIATE, the client MUST use a port number and address of all
   zeros.
```
Most socks server just ignore this destination, received on step 2.

Meteorite:
So, to track multiple destinations and multiple clients server must remember source IP and port number of incoming packet from client and associate them with UDP stream. When backwards traffic from destination is received it must be forwarded to client's IP:port, from where original forward UDP traffic was received by server.

So, if multiple clients want to send UDP packets to the same destination, they can be provided with the same port, but server will use their IPs and port numbers to distinguash between their UDP streams. If multiple clients want to send UDP packets to the same destination, server must send packets on their behalf from different ports to be able to distinguish backwards traffic.

So, UDP socks server can be thought of as a kind of NAT.

Meteorite:
As I already said in PR description, github.com/meteorite/socks5 library has 2 strategies for UDP ports handling:

- SingleUDPPortAssociate - opens one UDP port, when socks server starts and uses it for the lifetime of server to serve all clients (for chisel it means one UDP port per socks remote). Main benefits: 1) using less ports, 2) more firewall friendly, as port numbers can be made predictable. Main drawback: slightly poorer performance due to less parallel packets processing (some additional locks are needed to find UDP stream for each incoming packet)
- MultiUDPPortAssociate - opens separate UDP port for every client and closes it, when client closes corresponding TCP-session. Benefits and drawbacks are opposite to previous strategy.

For now, SingleUDPPortAssociate is used in PR, that’s why, we might do it a little bit more firewall friendly, if we want.

Meteorite:
> doesn’t this need to be a port range rather than a specific UDP port?

Single port per remote is enough for SingleUDPPortAssociate, but we can specify a range, if we want to switch to MultiUDPPortAssociate strategy or to allow searching for free port during initialization with SingleUDPPortAssociate.

Meteorite:
By the way, in this PR chisel creates one SSH channel per every UDP client app's IP:port pair of every socks remote. If client app's is sending UDP packets from the same IP:port to multiple destinations, then the same channel is reused for all such UDP streams. That's why each forward UDP packet is send via this channel with a header, indicating this packet destination, and each backward UDP packet is sent back via the channel with header, indicating remote party, from which this packet was received to transfer it to client app.

See: [here](https://github.com/jpillora/chisel/pull/250/files#diff-0ea63d9962e026f3e5c41776b7f919a770fc408d93438a741d4546e31c643e68R76) and [here](https://github.com/jpillora/chisel/pull/250/files#diff-0ea63d9962e026f3e5c41776b7f919a770fc408d93438a741d4546e31c643e68R130)

jpillora:
Hey guys, thanks for the summary. I like where the intuitive use of where just works™ - and this is an area where things can silently fail, so if we can stick to single-port associate and not support multi-port then that's great.

If we just do single-port, then socks server still need to be on the inbound side (`tunnel_in_*.go`)?

Meteorite:
> If we just do single-port, then socks server still need to be on the inbound side (`tunnel_in_*.go`)?

Yes, I think it is still far better to be on the inbound side, especially for forward remotes, as chisel server can work with multiple chisel clients simultaneously and each of chisel client can have multiple forward socks remotes with different UDP ports. If outbound side have to create UDP ASSOCIATE reply, then chisel server would need to know all those remotes' ports of all connected chisel clients and all connected chisel clients' IP addresses (as they must be reported to socks client app in UDP ASSOCIATE reply). Note, that now chisel server doesn't know chisel clients' inner IP addresses at all.

And, may be even more important is that all that ASSOCIATE handshake would have to be transferred via SSH via WS from inbound to outbound side and back, if socks server is on the outbound side. To my mind, it is very inefficient comparing with replying from inbound side itself. And forward UDP packets (from client app to destination) filtering functionality would also be located on outbound side, that would require to transfer each UDP packet from inbound to outbound side even if this packet must be dropped.

And, besides, having socks server on inbound side seems to be more flexible, as we could easily add some additional client traffic processing without additional passing inbound side's parameters to outbound side (i.e. without further chisel protocol changes).

jpillora:
Hey @aus do you have time to review this? Unfortunately, I do not at this stage - though if you're happy to merge - feel free to do so, I'll tag a new release once it's in :)

antifob:
Gentle bump :) I don't know what's left to do, but it's working for me.

Thanks for that awesome work @Meteorite

ip-rw:
@Meteorite I'm stealing your socks5 server for use in other projects :) many thanks.

I'm going to benchmark against the incumbent but this really looks worth merging...

ip-rw:
Have yet to figure out a decent way to speed test UDP performance making this a little redundant but:

proxychains iperf3 -N -c ping.online.net -p 5204 -P5 (via a chisel-udp-associate socks5 proxy to a machine in the same rack)
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
...
[SUM]   0.00-10.00  sec  1.08 GBytes   931 Mbits/sec   62             sender
[SUM]   0.00-10.00  sec  1.04 GBytes   895 Mbits/sec                  receiver

proxychains iperf3 -N -c ping.online.net -p 5204 -P5 -R (via a chisel-udp-associate socks5 proxy to a machine in the same rack)
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
...
[SUM]   0.00-10.00  sec   269 MBytes   225 Mbits/sec  574             sender
[SUM]   0.00-10.00  sec   221 MBytes   185 Mbits/sec                  receiver

proxychains iperf3 -N -c ping.online.net -p 5204 -P5 (via a chisel socks5 proxy to a machine in the same rack)
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
...
[SUM]   0.00-10.00  sec  1.09 GBytes   933 Mbits/sec   81             sender
[SUM]   0.00-10.00  sec  1.04 GBytes   894 Mbits/sec                  receiver

proxychains iperf3 -N -c ping.online.net -p 5204 -P5 -R (via a chisel socks5 proxy to a machine in the same rack)
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
...
[SUM]   0.00-10.00  sec   248 MBytes   208 Mbits/sec   67             sender
[SUM]   0.00-10.00  sec   201 MBytes   169 Mbits/sec                  receiver



I'm aware these benchmarks mean very little, really I should be running iperf3 on the chisel host machine... or doing this more than once.

Obviously I'm yet to test the UDP bits, I'll just make a thing to fling packets over several ports and see what happens... nothing off the shelf I can find to do it.

At a glance, for TCP it seems to perform well.

Meteorite:
@ip-rw , you are welcome! :)

ip-rw:
> @ip-rw , you are welcome! :)

i use this in another project and it works great :)

PR #255, opened 2021-03-21
Client Auth Options

Ronni3X:
I added more options for how the authorization information is provided through the client (the --auth option). You can provide the username and password through stdin by putting the string "stdin" in place of the <username:password>. It will prompt you for the username and the password, with the typed password not being echoed. To implement this I had to include the term library from the golang repo. This was useful for me in a lab environment, like Hack the box and OSCP, so that other users couldn't see the username and password in the process list output (ps) and then create connections back through my machine. I added the file and environment variable options for the same reason as the stdin option, but also for the lack of a full tty shell. After providing the input file or environment variable you can then remove the login information. The file and environment variable option also allow you to background the process when ran. To provide the file you include a greater than symbol followed by the file name (i.e. ">auth_file"). If the file is in the same path you can use only the name, otherwise you need to provide the relative or absolute path with the filename. To provide the environment variable you include an equal symbol followed by the environment variable name (i.e. "=auth"). I changed the help information and README to reflect these additions. I'm new to golang so the code might not be as efficient as it could be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,14 @@ $ chisel server --help
 
     --auth, An optional string representing a single user with full
     access, in the form of <user:pass>. It is equivalent to creating an
-    authfile with {"<user:pass>": [""]}. If unset, it will use the
-    environment variable AUTH.
+    authfile with {"<user:pass>": [""]}. To provide auth info securely,
+    use the string "stdin" instead of the normal "user:pass" string. If
+    a full tty shell is not present you can provide auth info as a file
+    or an environment variable. To provide a file, use a greater than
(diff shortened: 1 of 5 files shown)


haegar:
Moin,

Can't comment on the go code, but I like the principle of not having the authentication info visible in the process list.

Just some small comment:

"To provide the file you include a greater than symbol followed by the file name (i.e. ">auth_file")"
Here I think using the less than symbol "<" would make more sense - as this option here means "read from", not "write to", we should use the same as the shell and not the opposite.
Then if you make an error in your shell quoting you get maybe an error, but did not overwrite your auth data with the chisel output.

PR #258, opened 2021-03-25
Added socks5 auth and filters

marcino239:
Hi,

I've added two options:
- --socks5-auth - to specify socks5 credentials
- --socks5-allowed-hosts - to specify a simple filter

Hope you find those useful

Best Regards,
Marcino239

### DIFF
--- a/main.go
+++ b/main.go
@@ -342,6 +342,15 @@ var clientHelp = `
     the credentials inside the server's --authfile. defaults to the
     AUTH environment variable.
 
+    --socks5-auth, An optional username and password for socks5 server
+    in the form: "<user>:<pass>". These credentials will then be required
+    by Socks5 server when new client is connecting. Defaults to the
+    AUTH_SOCKS5 environment variable.
+
+    --socks5-allowed-hosts, An optional list of hosts (comma delimited) on
+    the client side which will be allowed to connect to by socks5 client.
+    Only FQDN values are accepted
+
     --keepalive, An optional keepalive interval. Since the underlying
     transport is HTTP, in many instances we'll be traversing through
     proxies, often these proxies will close idle connections. You must
@@ -400,6 +409,10 @@ func client(args []string) {
 	flags.StringVar(&config.TLS.Cert, "tls-cert", "", "")
 	flags.StringVar(&config.TLS.Key, "tls-key", "", "")
 	flags.Var(&headerFlags{config.Headers}, "header", "")
+
+	flags.StringVar(&config.Socks5Config.Auth, "socks5-auth", "", "")
+	socks5Allowedhosts := flags.String("socks5-allowed-hosts", "", "")
+
 	hostname := flags.String("hostname", "", "")
 	pid := flags.Bool("pid", false, "")
 	verbose := flags.Bool("v", false, "")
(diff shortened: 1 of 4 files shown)


share/tunnel/socks5.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (p *Socks5AllowedHosts) Allow(ctx context.Context, req *socks5.Request) (context.Context, bool) {
+	// only process FQDN
+	_, ok := p.AllowedHosts[req.DestAddr.FQDN]
+	return ctx, ok
+}
+
+//creates Socks5 server with optional authentication and host filter
+func CreateSocks5Server(sl *log.Logger, c Config) (*socks5.Server, error) {
```
jpillora, line 27: because it fatals, i think we can drop the return `error`

share/tunnel/socks5.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/armon/go-socks5"
+	"golang.org/x/net/context"
+)
+
+type Socks5Config struct {
+	Auth         string
+	AllowedHosts []string
+}
+
+type Socks5AllowedHosts struct {
```
jpillora, line 16: looks like we can unexport this one

share/tunnel/socks5.go
```diff
@@ -0,0 +1,51 @@
...
+
+//creates Socks5 server with optional authentication and host filter
+func CreateSocks5Server(sl *log.Logger, c Config) (*socks5.Server, error) {
+	socksServerConfig := socks5.Config{Logger: sl}
+	if c.Socks5Config.Auth != "" {
+		if !strings.Contains(c.Socks5Config.Auth, ":") {
+			sl.Fatal("Invalid format of socks5-auth. Missing colon separator")
+		}
+		creds := strings.Split(c.Socks5Config.Auth, ":")
+		cred := socks5.StaticCredentials{
```
jpillora, line 34: should also check for `len(creds) == 2`

share/tunnel/socks5.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+//creates Socks5 server with optional authentication and host filter
+func CreateSocks5Server(sl *log.Logger, c Config) (*socks5.Server, error) {
+	socksServerConfig := socks5.Config{Logger: sl}
+	if c.Socks5Config.Auth != "" {
+		if !strings.Contains(c.Socks5Config.Auth, ":") {
+			sl.Fatal("Invalid format of socks5-auth. Missing colon separator")
+		}
+		creds := strings.Split(c.Socks5Config.Auth, ":")
```
jpillora, line 33: can `SplitN 2` to allow for colons in the password

main.go
```diff
@@ -342,6 +342,15 @@ var clientHelp = `
...
 
+    --socks5-auth, An optional username and password for socks5 server
+    in the form: "<user>:<pass>". These credentials will then be required
+    by Socks5 server when new client is connecting. Defaults to the
+    AUTH_SOCKS5 environment variable.
+
+    --socks5-allowed-hosts, An optional list of hosts (comma delimited) on
+    the client side which will be allowed to connect to by socks5 client.
+    Only FQDN values are accepted
+
```
jpillora, line 353: i think this needs to be in the common section, since the socks server can run in both client and server modes

jpillora (commented):
thanks for the PR, sorry for the late review!

kblanckaert:
How does the authentication work? 
I've tested this (added --socks5-auth), but I'm still able to connect with and without passing credentials.
I'm sure the code is reached, because entering wrong credentials format (e.g no colon) does exit the client immediately.

Issue #266, opened 2021-04-15
Adding X-Forwarded-Host headers in reverse proxy


alejandroliu:
Can you add a `X-Forwarded-Host` header to reverse proxy functionality?

This can be used for the case that the backend web server has virtual hosts or to do http->https redirection tricks.

This can be achieved easily by adding in [server.go](https://github.com/jpillora/chisel/blob/92d90be68a989377daf61294ef7458612d10da8c/server/server.go#L105) after line 105:

```
r.Header.Add("X-Forwarded-Host", r.Host)
```

lwiechec:
that would be great !

PR #267, opened 2021-04-16
Adds X-Forwarded-Host header to reverse proxy (See #266)

alejandroliu:
Adds a X-Forwarded-Host header to the reverse proxy's original request.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -103,6 +103,7 @@ func NewServer(c *Config) (*Server, error) {
 		server.reverseProxy = httputil.NewSingleHostReverseProxy(u)
 		//always use proxy host
 		server.reverseProxy.Director = func(r *http.Request) {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -103,6 +103,7 @@ func NewServer(c *Config) (*Server, error) {
 		server.reverseProxy = httputil.NewSingleHostReverseProxy(u)
 		//always use proxy host
 		server.reverseProxy.Director = func(r *http.Request) {
+			r.Header.Add("X-Forwarded-Host", r.Host)
```
jpillora, line 106: this "leaks" information, it tells the backend server where the "frontend" is - probably not an issue in 95% of cases - though i think this needs to be user defined. maybe you could wrap with `if EnvBool("X_FORWARDED_HOST") { ... }` ?

Issue #278, opened 2021-05-18
proxy tunnel logger


zuzgon:
Hey, I found out a problem I have I can't change the logger values of the proxy tun (the one forked when creating a new chisel).
https://github.com/jpillora/chisel/blob/92d90be68a989377daf61294ef7458612d10da8c/share/tunnel/tunnel.go#L52
https://github.com/jpillora/chisel/blob/92d90be68a989377daf61294ef7458612d10da8c/client/client.go#L178, 
since there's no way in the config (and its defaulted to info+debug) to let the client know I don't want it to log Info\Debug messages I have no way to specify to the tunnel I don't want it to log Info\Debug messages.

I'm talking about these lines 
![photo_2021-05-18_16-38-56](https://user-images.githubusercontent.com/12097542/118661176-8e788400-b7f7-11eb-99ae-17b8fc3c5b0f.jpg)

PR #281, opened 2021-05-24
Providing chisel's client with a logger level

zuzgon:
Providing chisel's client with a log level to stop having this issues:
https://github.com/jpillora/chisel/issues/278

Another solution could be providing a `cio.Logger` from the caller.
Changing `cio.Logger` to be an interface and then we'll be able to provide with a logger, 
I could do that as well if you prefer.


client/client.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	Headers          http.Header
 	TLS              TLSConfig
 	DialContext      func(ctx context.Context, network, addr string) (net.Conn, error)
+	Verbose          bool
```
jpillora, line 45: should also be exposed on the CLI i thinnk but this an old PR so will just merge 👍

jpillora (approved):

Issue #327, opened 2022-01-18
Respond to /health and /version by request path rather than the whole url string


BigSully:
Request for `/health?_t=anything` won't return the health info,  neither would request for /version if there exists extra query string.

Can we route the /health and /version by request path instead of the whole url string? I can make a pull request if it's okay.

substitute the [switch r.URL.String()](https://github.com/jpillora/chisel/blob/v1.7.6/server/server_handler.go#L37) with `switch r.URL.Path`.

```
switch r.URL.Path {
case "/health":
	w.Write([]byte("OK\n"))
	return
case "/version":
	w.Write([]byte(chshare.BuildVersion))
	return
}
```

jpillora:
send a PR?

PR #328, opened 2022-01-19
Respond to /health and /version by request path rather than by the whole url string

BigSully:
[Respond to /health and /version by request path rather than by the whole url string](https://github.com/jpillora/chisel/issues/327)


jpillora (approved):
Thanks!

jpillora (approved):

PR #339, opened 2022-03-04
Disable the default tcp keepalive mechanism.

ximiximi1:
The default tcp keepalive mechanism is enabled default by golang, which will send tcp keepalive packet every 15s. So we get too many keepalive packets when captured by wireshark.
 
Since we already have keepalive mechanism in websocket, so disable the keepalive  mechanism of tcp connection is reasonable.

### DIFF
--- a/client/client_connect.go
+++ b/client/client_connect.go
@@ -97,7 +97,6 @@ func (c *Client) connectionOnce(ctx context.Context) (connected, retry bool, err
 	ulwsConn := wsConn.UnderlyingConn()
 	tcpConn, ok := ulwsConn.(*net.TCPConn)
 	if ok {
-		c.Debugf("SetKeepAlive false\n")
 		tcpConn.SetKeepAlive(false)
 	}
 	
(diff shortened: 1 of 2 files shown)


jpillora:
When I first tested this on heroku, and another PaaS, the TCP keep alive was only sent to their external gateway - however explicit websocket data makes it end-to-end. If there's a strong usecase to disable by default, happy to hear it though will close for now

jpillora:
Sorry I misread the PR! You're disabling the standard TCP keepalive, not chisels keepalive. I think this is okay actually...

ximiximi1:
> When I first tested this on heroku, and another PaaS, the TCP keep alive was only sent to their external gateway - however explicit websocket data makes it end-to-end. If there's a strong usecase to disable by default, happy to hear it though will close for now

Yes, the TCP keepalive packet was sent to the external gateway to keep the tcp connection alive. So this is useless and duplicated since we have keepalive mechanism in chisel which is end-to-end.
Because the chisel keepalive is enabled by default, so disable the default tcp keepalive is fine and the packet every 15s is too often.

Issue #341, opened 2022-03-12
Data race in Proxy.pipeRemote()


andres-portainer:
There is a data race in this line https://github.com/jpillora/chisel/blob/master/share/tunnel/tunnel_in_proxy.go#L125.

Here is a simple example that shows the problem:

```go
package main

import (
	"context"
	"log"
	"net/http"
	"time"

	chclient "github.com/jpillora/chisel/client"
	chserver "github.com/jpillora/chisel/server"
)

func main() {
	// Server
	srv, err := chserver.NewServer(&chserver.Config{
		Reverse: true,
	})
	if err != nil {
		log.Fatal(err)
	}

	if err := srv.Start("127.0.0.1", "55555"); err != nil {
		log.Fatal(err)
	}

	// Client
	cli, err := chclient.NewClient(&chclient.Config{
		Server:  "localhost:55555",
		Remotes: []string{"R:127.0.0.1:44444:33333"},
	})
	if err != nil {
		log.Fatal(err)
	}

	cli.Start(context.Background())

	for i := 0; i < 10; i++ {
		time.Sleep(1 * time.Second)
		http.Get("http://127.0.0.1:44444/")
	}
}
```

Which gives the following output while running it with the race detector enabled:

```
==================
WARNING: DATA RACE
Read at 0x00c00013f6a0 by goroutine 46:
  github.com/jpillora/chisel/share/tunnel.(*Proxy).pipeRemote()
      /home/user/gopath/pkg/mod/github.com/jpillora/chisel@v1.7.7/share/tunnel/tunnel_in_proxy.go:125 +0xf6
  github.com/jpillora/chisel/share/tunnel.(*Proxy).runTCP·dwrap·5()
      /home/user/gopath/pkg/mod/github.com/jpillora/chisel@v1.7.7/share/tunnel/tunnel_in_proxy.go:119 +0x71

Previous write at 0x00c00013f6a0 by goroutine 35:
  github.com/jpillora/chisel/share/tunnel.(*Proxy).pipeRemote()
      /home/user/gopath/pkg/mod/github.com/jpillora/chisel@v1.7.7/share/tunnel/tunnel_in_proxy.go:125 +0x114
  github.com/jpillora/chisel/share/tunnel.(*Proxy).runTCP·dwrap·5()
      /home/user/gopath/pkg/mod/github.com/jpillora/chisel@v1.7.7/share/tunnel/tunnel_in_proxy.go:119 +0x71

Goroutine 46 (running) created at:
[...]

PR #342, opened 2022-03-14
Add locking around the connection count to fix a data race.

andres-portainer:
Fixes https://github.com/jpillora/chisel/issues/341.


jpillora (approved):
Thanks, sorry for the slight delay :D

PR #344, opened 2022-03-17
[WIP] Add Support for HTTP/3 and QUIC

zandercodes:
I am in the process of adding `HTTP/3` and `QUIC` support for `chisel`. To make the transfer of `chisel` even faster than it already is.


jpillora:
Didn’t work?

zandercodes:
> Didn’t work?

The problem is that HTTP/3 uses UDP so it is relatively difficult to implement this because WebSocket is used as a base, which cannot be used under HTTP/3.

jpillora:
Oh yeah, you could skip the WS layer and just run the SSH layer on top of HTTP3?

gerardo-junior:
+1

Issue #340, opened 2022-03-09
chisel client - Avoiding plain-text password when http_proxy


yaakov-berkovitch:
Hello,

This is not a real issue or bug, but a question, even if I think it violates some basic security rules.

We are using chisel v1.7.6.
I'm wondering if there is a way to avoid passing a plain-text password in the chisel client command line. After starting the chisel client, anyone that list the running processes will see the command arguments and so the clear password.

I didn't find a way to hide this command or the password from users, so any help with this concern will be very helpful.

Thanks.

jpillora:
Good point, at this stage, no way to hide it from the process args (`ps aux` will display it)

Easiest change to fix this is to use an environment variable fallback for the `--proxy` option, like `CHISEL_HTTP_PROXY` and then define this env var in a systemd unit file (or similar). Feel free to raise a PR, will merge it

yaakov-berkovitch:
Thanks @jpillora, your workaround looks working and the password is hidden.

yaakov-berkovitch:
@tman204 your solution is also a good approach, and by reading the password from stdin will indeed make it invisible using ps.
I didn't find a PR for this change, why not creating a PR and having your changes part of releases ? 
@jpillora your advise would be very appreciated.

ayelet-ack:
@jpillora added a PR to fix this issue.
Can you please merge it? Thanks!

ayelet-ack:
@jpillora I opened a PR with a fix for this issue a while ago: https://github.com/jpillora/chisel/pull/361
could you please merge?

PR #361, opened 2022-05-25
avoid plain-text proxy password - read from terminal / input

ayelet-ack:
Following up on this issue: https://github.com/jpillora/chisel/issues/340
Added a way to avoid plain text proxy passwords: you can run chisel client with --proxy user:@<proxy_ip>, and you will be prompt for a password.
alternatively, you can pass the password to chisel this way using pipe.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -15,6 +15,9 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"syscall"
+	"os"
+	"bufio"
 
 	"github.com/gorilla/websocket"
 	chshare "github.com/jpillora/chisel/share"
@@ -27,6 +30,7 @@ import (
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/net/proxy"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/term"
 )
 
 //Config represents a client configuration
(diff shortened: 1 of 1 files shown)


ayelet-ack:
@jpillora could you merge this please? thanks!

client/client.go
```diff
@@ -261,6 +265,27 @@ func (c *Client) setProxy(u *url.URL, d *websocket.Dialer) error {
 	// CONNECT proxy
 	if !strings.HasPrefix(u.Scheme, "socks") {
 		d.Proxy = func(*http.Request) (*url.URL, error) {
+			if u.User != nil {
+				pass, _ := u.User.Password()
+				if pass == "" {
+					fmt.Print("*** Enter proxy password: ")
+					if term.IsTerminal(syscall.Stdin) {
```
jpillora, line 272: Does terminal case different from the read line case?

ayelet-ack: TBH I'm not sure what the difference is (not a GO programmer...).
But when I tried this with just terminal (i.e. just the "if" part w/o else), and I tried to pass the proxy password using pipe:
`echo some_password | chisel client ....`
I got this error:
`client: Connection error: inappropriate ioctl for device`
When I googled the issue I found this thread, and the suggested solution worked: https://github.com/golang/go/issues/19909#issuecomment-461311448
See also here:
https://github.com/golang/go/issues/19909#issuecomment-400979666

jpillora:
Thanks but i've decided to close this as this feature has the potential to conflict with the `stdio` remote. I'd consider loading client password from an env var or from a file. Comment here with your idea before you implement it though 👍

Issue #357, opened 2022-05-14
Need a way to override hardcoded max UDP message size of 9012


frs3:
The max UDP message size is currently hardcoded to 9012. We would like to increase it, since our max workload message size is greater.

Quick fix:

export CHISEL_UDP_MAX_SIZE=12000

share\tunnel\tunnel_out_ssh_udp.go:80
share\tunnel\tunnel_in_proxy_udp.go:83

const maxMTU = settings.EnvInt("UDP_MAX_SIZE", 9012)

or command line param e.g. --udp-max-size 12000

Masterxilo:
why don't you create a pull request?

It should probably read `const maxMTU = settings.EnvInt("CHISEL_UDP_MAX_SIZE", 9012)`.
Or as you said, use a command line param. Does chisel/go have a library (integrated) that allows both? Might be useful to be able to provide all commandline options via env. variables (or a config file), whatever is most convenient.

jpillora:
Yep happy to put max MTU behind a env setting

jpillora:
Plz raise PR

frs3:
PR https://github.com/jpillora/chisel/pull/367

PR #367, opened 2022-06-17
UDP buffer size override with CHISEL_UDP_MAX_SIZE environment variable

fsiegmund:
CHISEL_MAX_UDP_SIZE environment variable to allow override of default max size 9012.

#357


jpillora:
thanks!

PR #369, opened 2022-06-30
Ldap authentication

jdoucerain:
This PR is to allow LDAP authentication of chisel client.
The chisel server ldap configuration is maintained in a file pointed by the --ldap-config option
The ldap-config file structure is as follow:
BindDN for non anonymous search
BindPassword for non anonymous search
Url ldap server url
BaseDN for search
Filter ldap objects filter
IDMapTo ldap attribute used as id
CA file containing the PEM formatted certificates to trust when connecting to ldap server
Insecure true if we skip the certificates validation

The users are authentication against the ldap server.
The authorization pattern is unchanged (in authfile)
In the auth file we maintain the userid and the remotes. The password is left to an empty string


kblanckaert:
Hi,

Interesting feature! 
Would really like to see this added as well, but I wouldn't completely disable the password authentication functionality once LDAP configuration is in place. 

The documentation about how to use this could also be made more clear:
- An example of the _ldap.json_ configuration file
- The fact that the LDAP users have to be added to the _users.json_ file
- The fact that LDAP over port 389 doesn't seem to work (TLS issues when trying)

I've added some comments about the above mentioned improvements in your code.

README.md
```diff
@@ -175,10 +175,13 @@ $ chisel server --help
...
-    holding multiple PEM encode CA certificate bundle files, which is used to 
-    validate client connections. The provided CA certificates will be used 
-    instead of the system roots. This is commonly used to implement mutual-TLS. 
+    holding multiple PEM encode CA certificate bundle files, which is used to
+    validate client connections. The provided CA certificates will be used
+    instead of the system roots. This is commonly used to implement mutual-TLS.
 
+    --ldap-config, a path to a file containing the ldap authentication
+		configuration.
+
```
kblanckaert, line 184: {
  "BindDN": "CN=ldapUser,OU=Users,OU=example,DC=EXAMPLE,DC=COM",
  "BindPassword": "ldapUserPassword",
  "Url": "example.com:636",
  "BaseDN": "OU=Users,OU=example,DC=EXAMPLE,DC=COM",
  "Filter": "(&(objectClass=person)(objectClass=user))",
  "IDMapTo": "sAMAccountName",
  "CA": "",
  "Insecure": true
}

kblanckaert:
@jpillora Can you merge this please? Thanks!

jpillora (commented):
Looks reasonable from a quick glance

Would be great if we could flatten the 5 deep if-else blocks

Also what’s the binary size difference after importing ldap v3?

kblanckaert:
binary size difference: 12214.113 KB (master) -> 12397.106 KB (ldap-authentication)

server/server.go
```diff
@@ -177,10 +185,54 @@ func (s *Server) authUser(c ssh.ConnMetadata, password []byte) (*ssh.Permissions
...
+				s.Debugf("Could not authenticate user: %s", n)
+				return nil, errors.New("Invalid authentication for username: %s")
+			}
+		}
+	} else {
+	// User not found
+
+	s.Debugf("User not found: %s", n)
+	return nil, errors.New("Invalid authentication for username: %s")
 	}
```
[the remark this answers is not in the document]
jdoucerain: @kblanckaert
instead of
	if (!ldap_enabled && string(password) == "") {
		// Empty password

		s.Debugf("Empty passwords are not allowed. User = %s", n)
		return nil, errors.New("Invalid authentication for username: %s")
 I would rather do this
	if (string(password) == "") {
		// Empty password

		s.Debugf("Empty passwords are not allowed. User = %s", n)
		return nil, errors.New("Invalid authentication for username: %s")

giving that empty password is not allowed anyways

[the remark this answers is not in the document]
jdoucerain: @kblanckaert 
and the stanza 
	} else if (settings.LdapAuthUser(user,password,s.config.LdapConfig) == nil) {
		// LDAP authentication successful.
should be replaced by 
	} else if (ldap_enabled && settings.LdapAuthUser(user,password,s.config.LdapConfig) == nil) {
		// LDAP authentication successful.

to avoid failure

But to be honest I a not sure we are gaining any simplification vs what is currently proposed in the PR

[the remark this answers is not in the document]
kblanckaert: I've added that to avoid logging in with an empty password when there is no LDAP, but the same can be said for every placeholder character. Password authentication is still checked before LDAP, so you could log in by using that placeholder character.
Another option is checking LDAP authentication first (if enabled)

The 'ldap_enabled' should indeed be added to third _else if_ (especially when swapping the password & ldap auth statements)

jdoucerain:
@jpillora could you please have a look at the bit improved version of the password validation ?
Thanks

README.md
```diff
@@ -175,10 +175,15 @@ $ chisel server --help
...
-    instead of the system roots. This is commonly used to implement mutual-TLS. 
-
+    holding multiple PEM encode CA certificate bundle files, which is used to
+    validate client connections. The provided CA certificates will be used
+    instead of the system roots. This is commonly used to implement mutual-TLS.
+
+    --ldap-config, a path to a file containing the ldap authentication
+    configuration.
+    here is an example of an ldap-config file
+    { "BindDN": "CN=ldapUser,OU=Users,OU=example,DC=EXAMPLE,DC=COM", "BindPassword": "ldapUserPassword", "Url": "example.com:636", "BaseDN": "OU=Users,OU=example,DC=EXAMPLE,DC=COM", "Filter": "(&(objectClass=person)(objectClass=user))", "IDMapTo": "sAMAccountName", "CA": "", "Insecure": true }
```
jpillora, line 185: this `README.md` is generated using the command in the comment on L108 so this example will get replaced with whats in `main.go`

README.md
```diff
@@ -201,7 +206,7 @@ $ chisel server --help
 
 
 <!--tmpl,code=plain:echo "$ chisel client --help" && go run main.go client --help | sed 's#0.0.0-src (go1\..*)#X.Y.Z#' -->
-``` plain 
+``` plain
```
jpillora, line 209: these spacing edits come from your IDE i think,
can we revert them to reduce the size of the change set

easiest way is to execute the markdown template https://github.com/jpillora/md-tmpl

share/settings/ldap.go
```diff
@@ -0,0 +1,133 @@
+package settings
```
jpillora, line 1: can we `go fmt` this file?

share/settings/ldap.go
```diff
@@ -0,0 +1,133 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+
+	"github.com/go-ldap/ldap/v3"
+)
+
+//LdapConfig enables LDAP auth
+type LdapConfig struct {
```
jpillora, line 15: can we replace all capitalised acronym identifiers `LdapFooBar` with `LDAPFoobar` ?

server/server.go
```diff
@@ -177,10 +185,23 @@ func (s *Server) authUser(c ssh.ConnMetadata, password []byte) (*ssh.Permissions
...
+			// LDAP authentication successful
+		} else {
+			s.Debugf("unsuccessful authentication", n)
+			return nil, errors.New("Invalid authentication for username: %s")
+		}
+	} else {
+	// User not found or empty password
+
+	s.Debugf("User not found: %s or empty password", n)
+	return nil, errors.New("Invalid authentication for username: %s")
```
jpillora, line 202: i think we can further simplify

```
if !found {
  return nil, errors.New("user not found")
}
if string(password) == "" {
  return nil, errors.New("user password not set")
}
if (user.Pass == string(password)) {
	// local authentication successful
	return nil, nil
}
if (s.config.LdapConfigFile != "") {
	if err := settings.LdapAuthUser(user,password,s.config.LdapConfig); err != nil {
		return nil, fmt.Errorf("user ldap auth failed: %w", err)
	}
	// ldap authentication successful
	return nil, nil
}
return nil, errors.New("user auth failed")
```

share/settings/ldap.go
```diff
@@ -0,0 +1,133 @@
...
+	err = l.Bind(userdn,string(password[:]))
+	if err != nil {
+		return fmt.Errorf("Bad password for user %s",user.Name)
+	}
+
+	return nil
+}
+
+// Ldap Connection with TLS
+func ConnectTLS(ldapconfig LdapConfig) (*ldap.Conn, error) {
```
jpillora, line 72: can we unexport `ConnectTLS` and any other functions that arent used

main.go
```diff
@@ -159,9 +159,13 @@ var serverHelp = `
...
     --tls-ca, a path to a PEM encoded CA certificate bundle or a directory
-    holding multiple PEM encode CA certificate bundle files, which is used to 
-    validate client connections. The provided CA certificates will be used 
-    instead of the system roots. This is commonly used to implement mutual-TLS. 
+    holding multiple PEM encode CA certificate bundle files, which is used to
+    validate client connections. The provided CA certificates will be used
+    instead of the system roots. This is commonly used to implement mutual-TLS.
+
+    --ldap-config, a path to a file containing the ldap authentication
+		configuration.
```
jpillora, line 167: it's not super clear what `--ldap-config` does once set. maybe:

```
    --ldap-config, a path to a JSON configuration file, which defines settings used to
    connect to a remote LDAP server for authenticating users. once configured, user
    passwords will be validated against the configured LDAP server. here is an example
    JSON file: { "something": "something" }
```

server/server.go
```diff
@@ -177,10 +185,23 @@ func (s *Server) authUser(c ssh.ConnMetadata, password []byte) (*ssh.Permissions
 	// check the user exists and has matching password
 	n := c.User()
 	user, found := s.users.Get(n)
-	if !found || user.Pass != string(password) {
-		s.Debugf("Login failed for user: %s", n)
-		return nil, errors.New("Invalid authentication for username: %s")
+	if (found && string(password) != "") {
```
jpillora, line 188: although the users config file acts as a nice whitelist and access control, its not technically required - ldap can tell us both, if a user exists, and if the password is correct. should it work no users defined?

share/settings/ldap.go
jdoucerain: go fmt run as requested

share/settings/ldap.go
jdoucerain: changed as requested

server/server.go
jdoucerain: changed as recommended

main.go
jdoucerain: changed as proposed

server/server.go
jdoucerain: My opinion on this is the following:
if we want to use LDAP to only authenticate and forget about authorization, I think it should be explicitly mentioned with an additional flag to avoid confusion.
Now from the enterprise perspective (which was coming with the LDAP authentication) I am not seeing the only authentication and no authorization as a use-case and if I may I would deal with this as a separate change.

jdoucerain:
@jpillora thanks for the review you did and I tried to follow what you said.
Could you please review this change again ?
Thanks again

jdoucerain:
Hi @jpillora would you have some time to check if you could merge this branch ?

jdoucerain:
Hi @jpillora could you please check if you could merge this branch ? Thanks

jpillora:
Looks pretty good, will try take a look toknight

I think the server config should only have the ldap config, the path string should be in main (cli code) - settings can export a LoadLDAP(path) LDAPConfig which is used my main, and by server package consumers

I’m happy to do this change if you’d like

jdoucerain:
@jpillora please go ahead if you have some bandwidth. Thanks

jpillora:
looking at the is PR now. it appears you still have to define users in the --auth or --authfile, and the "local" password still works. ldap just adds an additional password that will now work. can you confirm this is correct and intended?

jpillora:
IMO, i think ldap should work in either one of two ways:

1. using ldap should be used if the local user is not found. so by default all user attempts use ldap. `config.AllowedUsers []string` could be added to whitelist users. maybe you could place remote configuration as a user attribute? otherwise all remotes are allowed
2. using ldap should force all "local" passwords to be blank. must be `--auth foo` if `--auth foo:bar` then fails to start. similarly with `--authfile`

jdoucerain:
> looking at the is PR now. it appears you still have to define users in the --auth or --authfile, and the "local" password still works. ldap just adds an additional password that will now work. can you confirm this is correct and intended?

jdoucerain:
Yes this is intentional. We have to cover case where the users could or could not be in the enterprise directory.

jdoucerain:
And the presence in the directory is not enough for us to allow anyone to use chisel. This is why we are maintaining the authfile for each individual. This is our use case.

jdoucerain:
I think using ldap as authentication only and applying a default parameterized remote is another use case to be developed.

jpillora:
gotcha, would you prefer option 2?

jdoucerain:
Yes I agree with option 2

jdoucerain:
@jpillora option 2 (using ldap should force all "local" passwords to be blank. must be --auth foo if --auth foo:bar then fails to start. similarly with --authfile) is implemented.

jpillora:
Thanks! Hopefully will get some more time soon

PR #371, opened 2022-07-10
README: remove -v from go install command

codenoid:
When I try the old command, I got:

```sh
$ go install github.com/jpillora/chisel@latest -v
go: -v: all arguments must have the same version (@latest)
```

remove the -v fix the problem


jpillora (approved):

jpillora:
Thanks! Strange that I had the -v at the end. Should be after install

PR #376, opened 2022-08-26
Set ServerName (SNI) to *hostname.

ip-rw:
Set ServerName (SNI) to *hostname. Useful for spoofing our way through restrictive gateways.


jpillora (approved):
Thanks @ip-rw !

Issue #394, opened 2022-11-23
build for armel


maurerr:
I would like to run chisel on an old Broadcom BCM4708A0 ARM EABI (https://wiki.debian.org/ArmEabiPort).
Is it possible to get a build?

managed to do it in the meantime (arm_v5)- I'll leave it here if other may find it useful https://www.dropbox.com/s/7j7q0ccs0ve03fk/chisel.zip?dl=1

Thanks

PR #395, opened 2022-11-23
add arm v5 builds

maurerr:
should fix https://github.com/jpillora/chisel/issues/394


jpillora:
Releasing as `v1.9.1` thanks :)

PR #403, opened 2023-02-01
Align Chisel server logging with client

benbek:
In a manner identical to #281 which makes it possible to set the client's logging level, this provides the option to set the logging level in the server.


jpillora:
I'd like to switch to go 1.21 `slog` and then add a new flag to set verbosity, and then `--verbose ` then equivalent is `--level DEBUG`

Issue #405, opened 2023-02-04
chisel crash randomly


magnww:
version: 1.8.1 linux amd64

chisel server -p 1111
chisel client 4.5.6.7:1111 1111:2222/udp

Client connect to server through a https proxy, and the network may not be stable.

```
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 13 [running]:
sync.(*WaitGroup).Wait(0x0?)
        sync/waitgroup.go:141 +0x85
github.com/jpillora/chisel/share/tunnel.(*waitGroup).Wait(...)
        github.com/jpillora/chisel/share/tunnel/wg.go:31
github.com/jpillora/chisel/share/tunnel.(*Tunnel).activatingConnWait.func1()
        github.com/jpillora/chisel/share/tunnel/tunnel.go:137 +0x2c
created by github.com/jpillora/chisel/share/tunnel.(*Tunnel).activatingConnWait
        github.com/jpillora/chisel/share/tunnel/tunnel.go:136 +0x85
```

jpillora:
The `WaitGroup` panic reported here was fixed in v1.11.7 (#586). v1.12.0-rc1 additionally hardens the UDP path (idle flow sweep) and reconnect behavior, both of which this setup exercises.

Please try out https://github.com/jpillora/chisel/releases/tag/v1.12.0-rc1 and report any issues in https://github.com/jpillora/chisel/issues/610

PR #406, opened 2023-02-05
Improve udp tunnel stability

magnww:
#405

### DIFF
--- a/share/tunnel/tunnel.go
+++ b/share/tunnel/tunnel.go
@@ -39,6 +39,7 @@ type Tunnel struct {
 	//ssh connection
 	activeConnMut  sync.RWMutex
 	activatingConn waitGroup
+	activatingConnMut sync.RWMutex
 	activeConn     ssh.Conn
 	//proxies
 	proxyCount int
@@ -134,7 +135,9 @@ func (t *Tunnel) getSSH(ctx context.Context) ssh.Conn {
 func (t *Tunnel) activatingConnWait() <-chan struct{} {
 	ch := make(chan struct{})
 	go func() {
+		t.activatingConnMut.Lock()
 		t.activatingConn.Wait()
+		t.activatingConnMut.Unlock()
 		close(ch)
 	}()
 	return ch
--- a/share/tunnel/tunnel_in_proxy_udp.go
+++ b/share/tunnel/tunnel_in_proxy_udp.go
@@ -121,6 +121,9 @@ func (u *udpListener) runOutbound(ctx context.Context) error {
 		//upsert ssh channel
 		uc, err := u.getUDPChan(ctx)
 		if err != nil {
+			if err.Error() == "ssh-conn nil" {
+				continue
+			}
 			if strings.HasSuffix(err.Error(), "EOF") {
 				continue
 			}


share/tunnel/tunnel.go
```diff
@@ -127,8 +127,19 @@ func (t *Tunnel) getSSH(ctx context.Context) ssh.Conn {
 		t.activeConnMut.RLock()
 		c := t.activeConn
 		t.activeConnMut.RUnlock()
-		return c
+	return c
+	}
+}
+
+func (t *Tunnel) getSSHNoWait(ctx context.Context) ssh.Conn {
```
jpillora, line 134: isn't `getSSH` equivalent when `c != nil` ?

magnww: Yes you are right.
Remove the wait may also cause high cpu usage, I will revert this change.

share/tunnel/tunnel_in_proxy_udp.go
```diff
@@ -100,15 +100,19 @@ func (u *udpListener) runInbound(ctx context.Context) error {
 			if strings.HasSuffix(err.Error(), "EOF") {
 				continue
 			}
-			return u.Errorf("inbound-udpchan: %w", err)
+			u.Errorf("inbound-udpchan: %w", err)
+			// wait for a short while for the ssh connection
+			time.Sleep(time.Duration(100) * time.Millisecond)
```
jpillora, line 105: instead of sleeping, can use channels to know when the ssh connection is ready?

magnww: The reason for adding sleep before is that, when the connection is lost, getSSH() will return nil immediately, and the runOutbound() loop will take up a lot of memory and cause the process to be killed.

jpillora: Ohh the loop spins? I haven’t seen that happen before though that sounds like a bug, where the real fix might be to  block with a waitgroup

magnww: After connection lost, `activatingConnWait()` cannot block, because `t.activatingConn.DoneAll()` is called.
Removing `DoneAll()` fixes the issue.

share/tunnel/tunnel.go
```diff
@@ -68,6 +68,10 @@ func New(c Config) *Tunnel {
 	return t
 }
 
+func (t *Tunnel) Close() {
+	t.activatingConn.DoneAll()
```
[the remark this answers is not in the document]
magnww: activatingConn is a waitGroup, it is value type and thread safe, so no need for lock and nil checks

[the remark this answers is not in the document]
jpillora: ah yea that's right, its a custom wait group (which might be a bad idea itself - anyway)

it should block while disconnected/connecting  and it should not-block while connected

DoneAll here would stop blocking, but Tunnel Done means closed?

how does this improve UDP? sorry, I think I'm missing something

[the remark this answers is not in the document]
magnww: In the original code, when the connection is lost, DoneAll in the BindSSH method will be called, which causes activatingConnWait() to not work and cannot block.

This leads to high resource usage of runOutbound in udpListener. Calling `u.getUDPChan(ctx)` will immediately gets nil instead of blocking.

`Close` is added to correctly release resources in server mode, because DoneAll was removed from BindSSH.

In client mode, there is no need to call DoneAll because the same tunnel is always used.

jpillora:
v1.12.0-rc1 improves UDP tunnel stability: idle over-cap flows are swept (no more wedging past 100 flows) and dead tunnel connections are detected via ping timeout.

Please try out https://github.com/jpillora/chisel/releases/tag/v1.12.0-rc1 and report any issues in https://github.com/jpillora/chisel/issues/610

PR #427, opened 2023-05-18
Upgrade to Go 1.20

cmenginnz:
Due to changes in the internal implementation of Go, upgrading from Go 1.19 (or earlier) to Go 1.20 will have two impacts on Chisel:
1. Given a key-seed, Go 1.19 and Go 1.20 will generate different private keys.
2. Given a key-seed, Go 1.20 will generate two different private keys.
Generating different private keys implies generating different fingerprints.

My software heavily relies on Chisel. There are hundreds of client nodes that store their own fingerprints. When upgrading to Go 1.20, the changed fingerprints will cause all client nodes to lose connections to the server.

This PR attempts to address the two issues above by:
1. Allows users to specify a private key file instead of generating it through a key-seed.
2. When ecdsa.GenerateKey tries to read 1 byte from the DetermRand reader, DetermRand skips moving to the next hash. In this way, it can neutralize the effects of MaybeReadRand() and always generates the same private key for a given key-seed.


main.go
```diff
@@ -109,6 +109,11 @@ var serverHelp = `
     of man-in-the-middle attacks (defaults to the CHISEL_KEY environment
     variable, otherwise a new key is generate each run).
 
+    --private-key-file, An optional path to a PEM-encoded TLS private key. When
```
jpillora, line 112:
```suggestion
    --private-key-file, An optional path to a PEM-encoded SSH private key. When 
```

jpillora (commented):
thanks! will likely merge this soon

cmenginnz:
jpillora, thanks for your review.

cmenginnz:
Hey @jpillora, I pushed a commit into this PR to fix the test failure.  Thanks.

jpillora:
Oh, I'm hoping we can use the same fingerprints - the standard lib explicitly says that it's not deterministic, so we might need to change the key gen library - and offer an easier way generate key files

jpillora:
based on this https://github.com/golang/go/issues/38548#issuecomment-617322694 Go is guarding against chisel's use case explicitly - bad decision by early on to have a `key seed` option though here we are

we have two options:

1. copy and paste the go 1.19 and earlier generate key algorithm out of the standard library and directly into chisel

2. I think from the next release, key seed won't generate a deterministic key - instead it'll exit with a message saying "use the server --keygen /file/path, followed by a server --keyfile /file/path" option which will output a generated key there

option 1 could be considered insecure, so I think if we should support both 1 and 2, and if we go with 1 it should print the deprecation message above and suggest 2

cmenginnz:
@jpillora yes, it's good if we can use the same fingerprint. I will try to support options 1 and 2.

Currently, in my PR, the argument name is "--private-key-file".  Could you confirm that you want to rename it to "--keyfile"? Thanks.

jpillora:
Yea I think keyfile matches keyseed? Is there a dash between I’m on mobile
atm

cmenginnz:
@jpillora, Following your comments, both options are implemented.

`chisel server --key=test123`
Start a chisel server that has the same fingerprint as Go1.19.

`chisel server --keygen=/tmp/key`
Generate an ssh key file.

`chisel server --keyfile=/tmp/key`
Start a Chisel server using the specified ssh key file.

jpillora:
Closed with #440

Issue #375, opened 2022-08-19
[Feature Request] Add framework to support multiple transport mechanisms for the outer tunnel


CCob:
Support other types of client/server transport mechanisms beyond HTTP.  This will open up the possibility of creating tunnels for standard mTLS, QUIC and even virtual channels within RDP and Citrix.

jpillora:
chisel relies on the SSH protocol, which it current runs ontop of TCP > HTTP > Websockets. Websockets is used to provide a generic `net.Conn`, so as long as we can get a `net.Conn`, we could use a different transport.

This is fairly large addition though, so if you or someone reading this wants to work on it. Please message here to chat about the proposed changes first

CCob:
Yea, that is what I was thinking, an abstraction around `net.Conn` for various transports.  My Go skills unfortunately are lacking and I wouldn't do it justice, hopefully somebody else who does that has an interest in this can pick it up.

PR #428, opened 2023-05-21
Add Support for HTTP/3 and WebTransport

lleyton:
As the title says. Still a work in progress, currently running into some weird QUIC/HTTP3 quirks which I'm having issues debugging. Besides that.. should almost be done.

Semi related to #375


jpillora:
Thank you :) definitely will consider this - on mobile haven’t looked at all the code yet though few Qs though

- how does the performance compare to the web socket transport? 1. Throughout? 2. With / without high packet loss? (I expect quic to perform better with a bad connection)
- I’m wondering if it should listen on the given port on both tcp (websockets) and Udp (quic) by default - ignore Udp listen failures. And then a flag to choose either/both explicitly - where failures are not ignored.
- I can it handle it on quic without the request path check? Makes it more transparent, and more friendly with the backend fallback flag

lleyton:
Thanks! As for your questions:

* I haven't run benchmarks yet, but I expect QUIC to perform better on flakier connections or where network congestion may be high.
* Hmm, currently it will automatically listen on UDP (QUIC) if a TLS certificate is set, as QUIC requires that. I could definitely add a flag to set the option explicitly though.
* I assume you're talking about the `/wt` handler? Hm, I could probably detect if it's a WebTransport connection by checking the headers. I could also check if the request came over UDP.

PR #440, opened 2023-08-13
Bump to Go 1.21

jpillora:
- Bump deps
- Bump go
- Key file/gen (thanks @cmenginnz)


jpillora:
Can you plz review @cmenginnz ?

client/client.go
```diff
@@ -105,7 +105,7 @@ func NewClient(c *Config) (*Client, error) {
 		tlsConfig: nil,
 	}
 	//set default log level
-	client.Logger.Info = c.Verbose
+	client.Logger.Info = true
```
jpillora, line 108: im reverting this, users who want to silence chisel can >/dev/null

jpillora:
I’m thinking of switching keygenfile to use plain base 64, because PEM encoding is multi line and adds an extra step when passing it inline. I’m thinking `ck-<n base64 chars>` which would also make it easier to identify its inline - rather than a file path

cmenginnz (approved):
LGTM

cmenginnz:
> I’m thinking of switching keygenfile to use plain base 64, because PEM encoding is multi line and adds an extra step when passing it inline. I’m thinking `ck-<n base64 chars>` which would also make it easier to identify its inline - rather than a file path

The `ssh.ParsePrivateKey(key)` function actually supports RSA (PKCS#1), PKCS#8, DSA (OpenSSL), and ECDSA private keys. If we use a private file format here, we will lose the support of these algorithms and the existing tools such as `openssl`. 
Please let me know your decision and I'm willing to continue patching for it.

jpillora:
I think we'd want the key parser to support both PEM and chisel-key format (`ck-<163-base64-chars>` drop the 2 equals)

cmenginnz:
@jpillora I made a PR(https://github.com/jpillora/chisel/pull/441) to go121 branch that implements:
1. `--genkey` generates a key file in CK format
2. `--keyfile` accepts:
2.1  a file in CK format
2.2 a file in PEM formet
2.3 an inline key in CK format

jpillora:
It took a while, but its in! Thanks for your help @cmenginnz 😄

jpillora:
`v1.9.0` releasing now https://github.com/jpillora/chisel/actions/runs/5908510975

PR #441, opened 2023-08-16
add supports for chisel-key format

cmenginnz:


share/ccrypto/keys_helpers.go
```diff
@@ -0,0 +1,96 @@
...
+	privateKey, err := chiselKey2PrivateKey(chiselKey)
+	if err == nil {
+		return privateKey2PEM(privateKey)
+	}
+
+	return nil, err
+}
+
+func IsChiselKey(chiselKey []byte) bool {
+	return string(chiselKey[:len(ChiselKeyPrefix)]) == ChiselKeyPrefix
```
jpillora, line 95: This can be strings.HasPrefix

also, are key lengths static? Can we also check total length here?

cmenginnz: Not sure if the length is static or not.

Below is the code that calculates the length. Hard to say it will be constant forever. 
`encodedPrivateKey := make([]byte, base64.RawStdEncoding.EncodedLen(len(b)))`

cmenginnz: BTW,  the length of `encodedPrivateKey` is 162, but you mentioned it's 163.

jpillora (approved):

PR #442, opened 2023-08-20
add timeout for keepalive reply

ximiximi1:
The sshConn.Send Request will send message and block wait the reply. 
The keepalive will not work in some special condition. For example, when the server use memory snapshot to hibernate and did not close the socket normally and there is a websocket middware like Cloudflare to proxy the websocket, when the server hibernate with memory snapshot, Cloudflare will not close the websocket connection during a very long time(this maybe a issue of Cloudflare),  in the mean while, the keepalive block wait the reply(there will not have a reply as server hibernated), as a result, the ssh connection will not close until the  Cloudflare close the websocket(very long time above 10min).
So I add a reply timeout for the keepalive. A timeout error will raise when wait the keepalive reply too long, and then close the ssh connetion.


jpillora:
Thanks this is a good change - will try get a sec to test it soon

linus-gates:
Hey, Is this going to merge anytime soon? Thanks!!

dedosmedia:
Please I need this feature in a future release! 🙏

jpillora:
Thanks for working on this! A fix for the blocking keepalive ping has been implemented based on the approach in #581 (ping raced against a timeout, with a CHISEL_PING_TIMEOUT override) and will ship in the next release. Closing as superseded.

PR #461, opened 2023-10-19
Update --keyfile documentation; remove reference to Base64

daethnir:
The --keyfile expects a file containing a chiselKey (ECDSA key starting with "ck-") y or the chiselKey itself as an argument.

Documentation indicated you could supplying it a base64 encoded version of chiselKey, but that's not what the code expects today at least.


daethnir:
Would fix https://github.com/jpillora/chisel/issues/498

PR #488, opened 2024-01-12
Modify keepAliveLoop to async for enabling timeout on no-response ping

naisy:
# Usage Environment:

- The chisel server operates with a fixed IP address.
- The chisel client frequently changes its IP address when using mobile phone networks.
- The chisel client uses both tunneling and reverse tunneling.

# Problem:

- In the keepAliveLoop, the ping transmission part does not have a timeout. Consequently, if there is no response, the process halts. This leads to the SSH port not being released.
- Although new clients with new IP addresses can send data to the server, reverse tunneling is limited to communication between one server port and one client. Therefore, when the client's IP changes, the server cannot send data to the new client. This issue persists until the chisel server is restarted.

# Solution:

- Modify the ping process to be asynchronous. If there is no response within the specified keepalive time, the established SSH connection will be closed.

### DIFF
--- a/share/tunnel/tunnel.go
+++ b/share/tunnel/tunnel.go
@@ -173,18 +173,44 @@ func (t *Tunnel) BindRemotes(ctx context.Context, remotes []*settings.Remote) er
 }
 
 func (t *Tunnel) keepAliveLoop(sshConn ssh.Conn) {
-	//ping forever
-	for {
-		time.Sleep(t.Config.KeepAlive)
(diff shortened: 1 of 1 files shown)


jpillora:
Thanks for working on this! A fix for the blocking keepalive ping has been implemented based on the approach in #581 (ping raced against a timeout, with a CHISEL_PING_TIMEOUT override) and will ship in the next release. Closing as superseded.

PR #537, opened 2024-10-24
Fix client backoff

fin3ss3g0d:
The code [here](https://github.com/jpillora/chisel/blob/ab8f06a83048dca0c24dc0b06932dc98df54e8b1/client/client_connect.go#L22C1-L22C55) fails to set a minimum value for the backoff based on [this](https://github.com/jpillora/backoff) project.

image

This means if the server is down, the backoff minimum for every chisel user is set to a 100ms value by default according to the screenshot above and will result in extremely short backoffs (default behavior is shown below).

image

Specifically, the sequence will always be 100ms, 200ms, 400ms, 800ms, and 1.6s (if the server is down). This technically gives the user about 2 seconds to get the server back up before the client exits entirely, and it is noisy over the wire to send this many packets this fast. I hope you can see how this can be problematic based on use case. The updated behavior with this commit can be seen below and gives users more time to get the server back up and running:

image

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -37,6 +37,7 @@ type Config struct {
 	MaxRetryCount    int
 	MinRetryInterval time.Duration
 	MaxRetryInterval time.Duration
+	RetryJitter      bool
 	Server           string
 	Proxy            string
 	Remotes          []string
--- a/client/client_connect.go
+++ b/client/client_connect.go
@@ -19,7 +19,9 @@ import (
 
 func (c *Client) connectionLoop(ctx context.Context) error {
 	//connection loop!
-	b := &backoff.Backoff{Max: c.config.MaxRetryInterval, Min: c.config.MinRetryInterval}
+	b := &backoff.Backoff{Min: c.config.MinRetryInterval,
+		Max:    c.config.MaxRetryInterval,
+		Jitter: c.config.RetryJitter}
 	for {
 		connected, err := c.connectionOnce(ctx)
 		//reset backoff after successful connections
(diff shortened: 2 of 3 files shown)


jpillora:
That’s strange, chisel client should retry forever…

fin3ss3g0d:
Respectfully, I never said retrying was the issue or that the client would stop retrying unintentionally... the title of the PR is `Fix client backoff` because of the way you have configured the backoff for the client. There should be more time in between each retry and the user should be able to control the time with a new flag...

You always set `Min` of the `Backoff` to 100ms [here](https://github.com/jpillora/chisel/blob/ab8f06a83048dca0c24dc0b06932dc98df54e8b1/client/client_connect.go#L22C1-L22C55) and don't give the user the option to set it, so all backoffs will be 100ms, 200ms, 400ms, 800ms, and 1.6s if the server goes down. Original:

image

After fix:

image

You can do your own testing to verify what I am reporting. Start a server, connect a client, kill the server with the client still connected and you'll see all backoffs will be 100ms, 200ms, 400ms, 800ms, and 1.6s with no ability to customize.

liojacqs:
If not exposed as a client flag, I believe enabling Jitter by default would be a good thing as well:
`b := &backoff.Backoff{Jitter: true, Max: c.config.MaxRetryInterval, Min: c.config.MinRetryInterval}
`

fin3ss3g0d:
> If not exposed as a client flag, I believe enabling Jitter by default would be a good thing as well: `b := &backoff.Backoff{Jitter: true, Max: c.config.MaxRetryInterval, Min: c.config.MinRetryInterval} `

I'll add this to another commit @liojacqs soon 😄

fin3ss3g0d:
@liojacqs added the jitter option as a bool flag. @jpillora any updated thoughts so far on this PR?

PR #546, opened 2025-01-25
remove unnecessary nil check

celestix:
Line 48 already ensures that the error is nil in the listener function's scope afterwards therefore the error check is redundant.


jpillora:
Thanks! The always-true err==nil removal has landed (with credit), alongside a fix moving the dead LetsEncrypt warning into the branch where it was intended. Ships next release.

PR #552, opened 2025-03-26
PROXY v2 support

### LINKED ISSUE #540: [FEAT] Add Proxy Protocol support
Would it be possible to add PROXY protocol support? This would allow visitors coming through the tunnel to be identified by their real IP on the backend. Currently i can only view the IP of the tunneling server.

lleyton:
Closes #540


venkatamutyala:
@jpillora any this could get merged into main and into the next release?

